PR #5, opened 2021-05-21
add `experiment opa` subcommand

jzelinskie:
This change adds an OPA subcommand that has authzed builtin functions added to the OPA runtime.

Example usage:

```
~/C/o/zed ❯❯❯ ./zed experiment opa eval 'authzed.check("user:fred", "reader", "document:firstdoc", "")'
{
  "result": [
    {
      "expressions": [
        {
          "value": true,
          "text": "authzed.check(\"user:fred\", \"reader\", \"document:firstdoc\", \"\")",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```


cmd/zed/experiment.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func RegisterAuthzedBuiltins(system string, client *authzed.Client) {
+	rego.RegisterBuiltin4(
+		&rego.Function{
+			Name:    "authzed.check",
+			Decl:    types.NewFunction(types.Args(types.S, types.S, types.S, types.S), types.B),
+			Memoize: false,
+		},
+		func(bctx rego.BuiltinContext, a, b, c, d *ast.Term) (*ast.Term, error) {
```
josephschorr, line 57: maybe change `a, b, c, d`?

cmd/zed/experiment.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+			if err := ast.As(c.Value, &objectStr); err != nil {
+				return nil, err
+			}
+			if err := ast.As(d.Value, &zedToken); err != nil {
+				return nil, err
+			}
+
+			var subjectNS, subjectID string
+			err := stringz.SplitExact(subjectStr, ":", &subjectNS, &subjectID)
```
josephschorr, line 74: This also needs to support the implicit `...`

cmd/zed/experiment.go
```diff
@@ -0,0 +1,115 @@
...
+			request := &v0.CheckRequest{
+				TestUserset: &v0.ObjectAndRelation{
+					Namespace: stringz.Join("/", system, objectNS),
+					ObjectId:  objectID,
+					Relation:  relation,
+				},
+				User: &v0.User{UserOneof: &v0.User_Userset{Userset: &v0.ObjectAndRelation{
+					Namespace: stringz.Join("/", system, subjectNS),
+					ObjectId:  subjectID,
+					Relation:  "...",
```
josephschorr, line 94: Make sure this is changed like the other locations

internal/storage/config.go
```diff
@@ -41,7 +41,7 @@ func CurrentToken(cs ConfigStore, ts TokenStore) (Token, error) {
 		return Token{}, err
 	}
 
-	return ts.Get(cfg.CurrentToken, false)
+	return ts.Get(cfg.CurrentToken, true)
```
josephschorr, line 44: Why this change?

cmd/zed/experiment.go
jzelinskie: This was the style used in the example, but I agree it's not very useful.

internal/storage/config.go
jzelinskie: It's not used anywhere anymore. The only time we use --reveal-tokens is on list.

josephschorr (approved):
LGTM

Issue #21, opened 2021-08-31
PrintTable ignores io.Writer argument


njhale:
The `printers.PrintTable` function [ignores its `io.Writer` argument](https://github.com/authzed/zed/blob/838b4a25a7863d0c768d1c2812b742cd57ff12b2/internal/printers/table.go#L12), always writing to `os.Stdout` instead.

It doesn't seem to be causing any issues at the moment -- all call sites pass `os.Stdout` anyway -- but it didn't look like it was intentional either.

njhale:
A fix was merged.

Thanks @jzelinskie!

PR #22, opened 2021-08-31
fix(printers): print to target writer

njhale:
The `PrintTable` function ignores its `io.Writer` argument, always printing to `os.Stdout` instead.

Fix this by using the argument.

Addresses #21


jzelinskie (approved):
Thanks for the report and fix! LGTM

PR #25, opened 2021-09-24
Add config for github releases

ecordell:
Not totally sure that the gha release creation plays nicely with the goreleaser release creation; I'll need to do some testing on this branch to be sure.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,58 @@
...
+    needs: "create"
+    runs-on: "macos-latest"
+    steps:
+      - uses: "actions/checkout@v2"
+      - uses: "actions/setup-go@v2"
+        with:
+          go-version: "^1.17"
+      - name: "build"
+        run: |
+          go build -o darwin-zed /cmd/zed
```
jzelinskie, line 31: `./cmd/zed` or else this will try to build from the root directory

.github/workflows/release.yaml
```diff
@@ -0,0 +1,58 @@
...
+    runs-on: "ubuntu-latest"
+    outputs:
+      upload_url: "${{ steps.release.outputs.upload_url }}"
+    steps:
+      - uses: "actions/create-release@v1"
+        id: "release"
+        env:
+          GITHUB_TOKEN: "${{ github.token }}"
+        with:
+          draft: "true"
```
jzelinskie, line 17: is this supposed to be a bool or a string?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,58 @@
...
+          tag_name: "${{ github.ref }}"
+          release_name: "${{ github.ref }}"
+
+  build-mac:
```
jzelinskie, line 21: do we need to make this job depend on goreleaser?

.goreleaser.yml
```diff
@@ -0,0 +1,81 @@
+before:
+  hooks: ["go mod tidy"]
```
jzelinskie, line 2: why does this need mod tidy beforehand? we should be enforcing that for PRs

.goreleaser.yml
```diff
@@ -0,0 +1,81 @@
...
+    license: "Apache 2.0"
+    epoch: 0
+    formats:
+      - "apk"
+      - "deb"
+      - "rpm"
+
+dockers:
+  - image_templates:
+      - "quay.io/authzed/zed:{{ .Version }}-amd64"
```
jzelinskie, line 48: Does this not support manifest lists?

.goreleaser.yml
```diff
@@ -0,0 +1,81 @@
...
+  - image_templates:
+      - "quay.io/authzed/zed:{{ .Version }}-arm64"
+    ids: ["linux-arm64"]
+    goos: "linux"
+    goarch: "arm64"
+    use: "buildx"
+    build_flag_templates:
+      - "--platform=linux/arm64"
+
+docker_manifests:
```
jzelinskie, line 64: does this produce a manifest list?

.github/workflows/release.yaml
ecordell: whoops, nice catch

.github/workflows/release.yaml
ecordell: It depends on `create`, which is all it needs to ensure the release is there for it to upload.

The `release` step in goreleaser also will create a release, but it [only creates](https://github.com/goreleaser/goreleaser/blob/61bead8989ea95dc7168efb58874ad07527071cb/internal/client/github.go#L128) if it doesn't exist, so I think these will play nicely with each other

.goreleaser.yml
ecordell: yep, out of the images built in the docker step

jzelinskie (approved):

PR #29, opened 2021-09-28
use goreleaser to publish docker images

ecordell:
goreleaser expects to use the binaries it already built

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,6 @@ COPY . /go/src/zed
 RUN go mod download
 RUN go install ./cmd/zed
 
-FROM alpine:3.13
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -8,4 +8,6 @@ RUN go install ./cmd/zed
 FROM alpine:3.13
 RUN apk --no-cache add ca-certificates
 COPY --from=build /go/bin/* /usr/local/bin/
+FROM gcr.io/distroless/base
+COPY zed /
```
jzelinskie, line 12: Let's replace the existing alpine with distroless, just bring over the copy command.

jzelinskie (approved):

PR #35, opened 2021-09-29
move mac build into dist

ecordell:
dist is gitignored, so goreleaser won't complain about a dirty tree

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,10 +17,10 @@ jobs:
         run: |
           go build -o zed ./cmd/zed
           tar -cvf zed_${{ steps.get_version.outputs.version-without-v }}_darwin_amd64.tar.gz ./zed LICENSE README.md
-      - uses: actions/upload-artifact@v2
+      - uses: "actions/upload-artifact@v2"
         with:
-          name: darwin_zed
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -35,12 +35,14 @@ jobs:
...
+        run: |
+          mkdir dist
         uses: "battila7/get-version-action@v2"
-      - uses: actions/download-artifact@v2
+      - uses: "actions/download-artifact@v2"
         with:
-          name: darwin_zed
-          path: zed_${{ steps.get_version.outputs.version-without-v }}_darwin_amd64.tar.gz
+          name: "darwin_zed"
+          path: "dist/zed_${{ steps.get_version.outputs.version-without-v }}_darwin_amd64.tar.gz"
```
jzelinskie, line 45: why not use $TEMPDIR?

ecordell: the archive needs to be moved into a place that's gitignored, or goreleaser will refuse to run (dist is gitignored and is the directory goreleaser puts its output archives too)

ecordell: see: https://github.com/authzed/zed/runs/3739410030?check_suite_focus=true

jakedt (approved):

PR #62, opened 2021-10-29
Add bulk-delete command for relationships in zed

josephschorr:


jzelinskie:
We discussed turning the subject filter into a flag for both this and read before merging this PR.

josephschorr:
@jzelinskie Added

cmd/zed/relationship.go
```diff
@@ -70,25 +77,121 @@ var readCmd = &cobra.Command{
 	RunE:              readRelationships,
 }
 
-func readRelationships(cmd *cobra.Command, args []string) error {
+var bulkDeleteCmd = &cobra.Command{
+	Use:               "bulk-delete <resource_type:optional_resource_id> <optional_relation> <optional_subject_type:optional_subject_id#optional_subject_relation>",
```
jzelinskie, line 81: `<optional_subject_type:optional_subject_id#optional_subject_relation>` should not be on the Use here and should be instead in the description of the flag

jzelinskie: I am also not sold on using the `#` syntax

jzelinskie: I also think this should just be called delete to have symmetry with the API

cmd/zed/relationship.go
```diff
@@ -70,25 +77,121 @@ var readCmd = &cobra.Command{
...
+
+		reader := bufio.NewReader(os.Stdin)
+		response, err := reader.ReadString('\n')
+		if err != nil {
+			return err
+		}
+
+		if strings.ToLower(strings.TrimSpace(response)) != "y" {
+			return nil
+		}
```
jzelinskie, line 157: We should use github.com/AlecAivazis/survey/v2 instead of the standard library for a good user experience if we're going to have interactive elements

cmd/zed/relationship.go
```diff
@@ -70,25 +77,121 @@ var readCmd = &cobra.Command{
...
+				return err
+			}
+
+			counter++
+			if counter > 1000 {
+				break
+			}
+		}
+
+		if counter == 0 {
```
jzelinskie, line 140: force should probably only wrap the interactive part of this code

cmd/zed/relationship.go
```diff
@@ -70,25 +77,121 @@ var readCmd = &cobra.Command{
...
+	if err != nil {
+		return err
+	}
+
+	request, err := buildReadRequest(cmd, args)
+	if err != nil {
+		return err
+	}
+
+	if !cobrautil.MustGetBool(cmd, "force") {
```
jzelinskie, line 111: maybe we should have a different flag for skipping the read preview
`estimate-count=false` or something

cmd/zed/relationship.go
josephschorr: We already have a delete

cmd/zed/relationship.go
josephschorr: The explicit goal of force was to skip the read as well

cmd/zed/relationship.go
josephschorr: The `#` is from read

cmd/zed/relationship.go
josephschorr: Do you think a prompt without the count is useful though?

jzelinskie: Yes, if I know it's huge.

cmd/zed/relationship.go
```diff
@@ -70,25 +76,124 @@ var readCmd = &cobra.Command{
...
+				break
+			}
+
+			if err != nil {
+				return err
+			}
+
+			counter++
+			if counter > 1000 {
+				break
```
jzelinskie, line 135: Is there something you should to do close the response stream here?

cmd/zed/relationship.go
```diff
@@ -70,25 +76,124 @@ var readCmd = &cobra.Command{
...
+		if counter > 1000 {
+			message = "Will delete 1000+ relationships. Continue?"
+		}
+
+		prompt := &survey.Confirm{
+			Message: message,
+		}
+		survey.AskOne(prompt, &response)
+		if !response {
+			return nil
```
jzelinskie, line 158: I feel like this should probably exit 1 rather than 0

cmd/zed/relationship.go
```diff
@@ -70,25 +76,124 @@ var readCmd = &cobra.Command{
...
+		response := false
+		message := fmt.Sprintf("Will delete %d relationships. Continue?", counter)
+		if counter > 1000 {
+			message = "Will delete 1000+ relationships. Continue?"
+		}
+
+		prompt := &survey.Confirm{
+			Message: message,
+		}
+		survey.AskOne(prompt, &response)
```
jzelinskie, line 156: style nitpick:

```go
		message := fmt.Sprintf("Will delete %d relationships. Continue?", counter)
		if counter > 1000 {
			message = "Will delete 1000+ relationships. Continue?"
		}

		response := false
		survey.AskOne(&survey.Confirm{
			Message: message,
		}, &response)
```

cmd/zed/relationship.go
jzelinskie: The first comment still needs to be addressed.

`<optional_subject_type:optional_subject_id#optional_subject_relation>` should be in the description of the flag and not in `Use` for positional arguments.

josephschorr: Its literally a copy-paste from the read command; should I fix it there as well?

cmd/zed/relationship.go
josephschorr: I'll add it, but I don't see many using it

cmd/zed/relationship.go
```diff
@@ -70,25 +76,124 @@ var readCmd = &cobra.Command{
...
+			}
+		}
+
+		if counter == 0 {
+			fmt.Println("No matching relationships found")
+			return nil
+		} else if counter > 1000 {
+			fmt.Print(" [y/N]: ")
+		} else {
+		}
```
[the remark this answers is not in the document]
josephschorr: Whoops. Removed

cmd/zed/relationship.go
josephschorr: Added a cancel context

josephschorr:
Updated

cmd/zed/relationship.go
jzelinskie: yep!

cmd/zed/relationship.go
jzelinskie: no need for return after an exit 1

jzelinskie: ok we discussed a better way to do the error handling here

cmd/zed/relationship.go
jzelinskie: ok we're supporting this positional and the flag, and failing now if both are provided

jzelinskie (approved):

PR #64, opened 2021-11-10
gomod: bump cobrautil to v0.0.7

jzelinskie:


ecordell (approved):
LGTM

We can follow up with a fix for gofumport's deprecation

PR #70, opened 2022-01-11
Log dispatch metadata

jzelinskie:
This allows callers to inspect how expensive their requests were to the backend.


jzelinskie:
Actually, I might be able to implement the trailer logging as a grpc.CallOption which would make this far less verbose.

jakedt (approved):

PR #71, opened 2022-01-12
switch 99designs/keyring to zalando/go-keyring

### LINKED ISSUE #67: Building on macOS emits a deprecation warning
mac OS version: 11.6.2

on build of cmd i get:

``` 

# github.com/keybase/go-keychain
cgo-gcc-prolog:203:11: warning: 'SecTrustedApplicationCreateFromPath' is deprecated: first deprecated in macOS 10.15 - No longer supported [-Wdeprecated-declarations]
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecTrustedApplication.h:59:10: note: 'SecTrustedApplicationCreateFromPath' has been explicitly marked deprecated here

```

Maybe we still need to have this to support old versions of Mac OS ?

ecordell:
I tested locally that this can access existing entries on the keychain (on macos). Before we merge it's probably worth reviewing the differences in the two libraries.

Fixes #67

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/authzed/zed
 go 1.16
 
 require (
-	github.com/99designs/keyring v1.1.6
 	github.com/AlecAivazis/survey/v2 v2.3.2
 	github.com/TylerBrock/colorjson v0.0.0-20200706003622-8a50f05110d2
 	github.com/authzed/authzed-go v0.3.0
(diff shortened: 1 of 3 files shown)


jzelinskie:
We technically rely on a bit more behavior out of this library than Zalando's library or any of the alternatives support.
The 99designs library supports a few more platforms and falls back to an encrypted file if none of them are available.

Ideally we don't lose those platforms, but we could also just implement the fallback behavior and call that good enough.

PR #76, opened 2022-01-17
cmd: trailers only log debug, add tracing

jzelinskie:
This also updates authzed-go to use the default of 0 instead of -1 when trailers are not present.


ecordell (approved):
LGTM

PR #93, opened 2022-03-02
Have `zed import` automatically handle prefixing when applicable

### LINKED ISSUE #81: Support prefix rewrites for zed import just like the playground does
When one edits a schema in the playground then imports it back into a permissions system via the "import" button, the playground automatically prepends the permission system's prefix to each reference in the schema.

However, that rewrite doesn't happen on `zed import` operations, making it difficult to share a common schema file with multiple permission systems; each reference in the schema has to be manually modified to include the prefix for the destination permission system.

This is a request to either:
- fetch the prefix from the destination permissions system and then automatically prepend the prefix to each reference, just like the playground (ideal)
- pass in the prefix via a command line option (not as ideal)

josephschorr:
Following this change, `zed import` now supports two modes for handling prefixes:

1) A new CLI flag `--schema-definition-prefix` which will automatically rewrite the schema and test relationships with that prefix
2) If not specified, `zed import` will first call `ReadSchema` for the permissions system and, if a prefix is found, use that prefix automatically

Fixes #81

### DIFF
--- a/cmd/zed/import.go
+++ b/cmd/zed/import.go
@@ -9,10 +9,16 @@ import (
 
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
 	"github.com/authzed/authzed-go/v1"
+	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
+	"github.com/authzed/spicedb/pkg/schemadsl/generator"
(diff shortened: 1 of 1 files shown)


ecordell (approved):
LGTM

Do we want to do this for schema write (and maybe read) as well?

PR #115, opened 2022-04-28
Store insecure in the context and use automatically if present

### LINKED ISSUE #63: Zed context should store whether the context is insecure
Right now the context stores everything I need to talk to Authzed.com or a local SpiceDB except whether that connection should use TLS or not. This means that every zed request to SpiceDB needs the `--insecure` flag. This flag should be stored along with the rest of the context because it doesn't regularly change for a context.

josephschorr:
Fixes #63

### DIFF
--- a/cmd/zed/context.go
+++ b/cmd/zed/context.go
@@ -76,8 +76,12 @@ func contextListCmdFunc(cmd *cobra.Command, args []string) error {
 		}
 		secret := token.APIToken
 		if !cobrautil.MustGetBool(cmd, "reveal-tokens") {
-			prefix, _ := token.SplitAPIToken()
-			secret = stringz.Join("_", prefix, "<redacted>")
+			secret = token.Redacted()
+		}
+
+		insecureStr := ""
+		if token.IsInsecure() {
+			insecureStr = "   ✓    "
 		}
 
 		rows = append(rows, []string{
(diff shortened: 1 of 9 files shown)


cmd/zed/context.go
```diff
@@ -77,18 +77,33 @@ func contextListCmdFunc(cmd *cobra.Command, args []string) error {
...
+				secret = "<redacted>"
+			} else {
+				secret = stringz.Join("_", prefix, "<redacted>")
+			}
+		}
+
+		insecure := false
+		if token.Insecure != nil {
+			insecure = *(token.Insecure)
+		}
```
jzelinskie, line 90: Also might want to make this a method on the token

cmd/zed/context.go
```diff
@@ -77,18 +77,33 @@ func contextListCmdFunc(cmd *cobra.Command, args []string) error {
...
+
+		insecureStr := ""
+		if insecure {
+			insecureStr = "   ✓    "
```
jzelinskie, line 94: We should make this a constant now that two things are using it

cmd/zed/main.go
```diff
@@ -36,16 +36,16 @@ func defaultStorage() (storage.ConfigStore, storage.SecretStore) {
...
 }
 
-func dialOptsFromFlags(cmd *cobra.Command, token string) []grpc.DialOption {
+func dialOptsFromFlags(cmd *cobra.Command, token storage.Token) []grpc.DialOption {
 	opts := []grpc.DialOption{
 		grpc.WithUnaryInterceptor(zgrpcutil.LogDispatchTrailers),
 	}
 
-	if cobrautil.MustGetBool(cmd, "insecure") {
+	if cobrautil.MustGetBool(cmd, "insecure") || (token.Insecure != nil && *token.Insecure) {
```
jzelinskie, line 44: if it was a method this would just be `|| token.Insecure()`

jzelinskie (changes requested):

cmd/zed/context.go
josephschorr: They are actually slightly different in length

josephschorr:
Updated

jzelinskie (approved):

Issue #94, opened 2022-03-03
Support large, batch importing of relationships


josephschorr:
`zed import` currently constructs a single `WriteRelationships` call, which will be exceeded after ~5000 relationships. We should add support for batching the relationships to be imported into chunks, and executing those chunks in parallel.

As mentioned in Discord (https://discord.com/channels/844600078504951838/844600078948630559/949071336574181457), there are a number of issues to address:

> 1. The gRPC server has a limit on message sizes. This can easily be solved by batching the requests
> 2. Batching is tricky, serially executing each batch works to an extent, but it's not very scalable. This could be improved by executing each batch in a goroutine. Of course, this comes with some risk, as there's roughly a 9k batch size limit (see 3). If a zed import is trying to import 6 million+ rows, that's about 650 connections each trying to shove through 9k tuples.
> 3. Postgres and MySQL both have a limit on how many placeholders you can have in a single query. This appears to be 65535 for Postgres and MySQL (https://stackoverflow.com/a/49379324, https://stackoverflow.com/a/24447922). This roughly translating to a maximum of 9362 relationship tuple writes (assuming each is going to require 7 placeholders). I didn't initially hit this limitation because I was testing with an in memory database.

We'll need to see a reasonable limit on the number of parallelized write requests, have a progress bar for display, etc

PR #95, opened 2022-03-03
Add basic batching to zed import

bryanhuhta:
This PR adds basic batching support for the `import` command, allowing `zed` to process large relationship files. Previously, `zed import` would tip over with ~9k relationship tuples (this is dependent on a few factors, like which DB driver is being used).

The general strategy presented here is to load all relationship tuples into memory (this behavior is unchanged), then in batches of 5000, serially import the tuples. This is by no means the most performant batching implementation, but this change will at least allow `zed` to import arbitrarily large files (provided the entire file can be initially stored in memory).

Related: https://github.com/authzed/zed/issues/94

PR #120, opened 2022-05-11
Add batching functionality

bryanhuhta:
supersedes https://github.com/authzed/zed/pull/95
closes https://github.com/authzed/zed/issues/94

This PR implements a more testable variant of batching than the one presented in https://github.com/authzed/zed/pull/95. The default batch size is 1000 records with 1 worker. A single worker renders the import synchronous, making parallel processing opt-in.

### DIFF
--- a/cmd/zed/import.go
+++ b/cmd/zed/import.go
@@ -15,11 +15,14 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/authzed/zed/internal/decode"
+	"github.com/authzed/zed/internal/grpcutil"
 	"github.com/authzed/zed/internal/storage"
 )
 
 func registerImportCmd(rootCmd *cobra.Command) {
 	rootCmd.AddCommand(importCmd)
+	importCmd.Flags().Int("batch-size", 1000, "import batch size")
+	importCmd.Flags().Int("workers", 1, "number of concurrent batching workers")
 	importCmd.Flags().Bool("schema", true, "import schema")
 	importCmd.Flags().Bool("relationships", true, "import relationships")
 	importCmd.Flags().String("schema-definition-prefix", "", "prefix to add to the schema's definition(s) before importing")
@@ -47,7 +50,7 @@ var importCmd = &cobra.Command{
 	From a local file (no prefix):
 		zed import authzed-x7izWU8_2Gw3.yaml
 
-	Only schema: 
+	Only schema:
 		zed import --relationships=false file:///Users/zed/Downloads/authzed-x7izWU8_2Gw3.yaml
 
 	Only relationships:
@@ -104,7 +107,9 @@ func importCmdFunc(cmd *cobra.Command, args []string) error {
 	}
 
 	if cobrautil.MustGetBool(cmd, "relationships") {
-		if err := importRelationships(client, p.Relationships, prefix); err != nil {
+		batchSize := cobrautil.MustGetInt(cmd, "batch-size")
+		workers := cobrautil.MustGetInt(cmd, "workers")
+		if err := importRelationships(client, p.Relationships, prefix, batchSize, workers); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


internal/grpcutil/batch.go
```diff
@@ -0,0 +1,85 @@
...
+			err = multierror.Append(err, e)
+		}
+
+		if finishedBatches == totalBatches {
+			close(errCh)
+		}
+	}
+
+	return err.ErrorOrNil()
+}
```
jzelinskie, line 85: It's a bit of personal preference, but I like to follow the lead for the x/sync packages when it comes to implementing concurrency patterns.

You can find a WorkerPool example here: https://pkg.go.dev/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/semaphore

I haven't tested this, but something along the lines of this would be preferable and wouldn't add any new dependencies:

```go
func min(x, y int) int {
	if x <= y {
		return x
	}
	return y
}

func ConcurrentBatch(ctx context.Context, n, batchSize, maxWorkers int, each EachFunc) error {
	if n < 0 {
		return errors.New("cannot batch items of length < 0")
	} else if n == 0 {
		return nil // Batching zero items is a noop.
	}

	if batchSize < 1 {
		return errors.New("cannot batch items with batch size < 1")
	}

	if maxWorkers == 0 {
		maxWorkers = runtime.GOMAXPROCS(0)
	}

	sem := semaphore.NewWeighted(int64(maxWorkers))
	g, ctx := errgroup.WithContext(ctx)
	numBatches := (n + batchSize - 1) / batchSize
	for i := 0; i < numBatches; i++ {
		if err := sem.Acquire(ctx, 1); err != nil {
			return fmt.Errorf("failed to acquire semaphore for batch number %d: %w", i, err)
		}
		g.Go(func() error {
			defer sem.Release(1)
			batchNum := i // Properly copy i
			start := i * batchSize
			end := min(start+batchSize, n)
			return each(batchNum, start, end)
		})
	}
	return g.Wait()
}
```

The reason I use these primitives instead is because channels actually have a ton of edge cases. You can see some in this paper that might surprise you, and I've found others in production systems: https://cseweb.ucsd.edu/~yiying/GoStudy-ASPLOS19.pdf

jzelinskie (changes requested):

internal/grpcutil/batch.go
bryanhuhta: That paper is an extremely interesting read with a few surprising "gotchas" when it comes to channels. I'll adopt the `WorkerPool` approach. I also like the name being changed to `ConcurrentBatch`, indicating that things happen concurrently under the hood.

cmd/zed/import.go
```diff
@@ -164,13 +169,24 @@ func importRelationships(client *authzed.Client, relationships string, definitio
...
+	log.Info().
+		Int("batch_size", batchSize).
+		Int("workers", workers).
+		Int("count", len(relationshipUpdates)).
+		Msg("importing relationships")
 
-	if _, err := client.WriteRelationships(context.Background(), request); err != nil {
-		return err
-	}
+	err := grpcutil.ConcurrentBatch(context.Background(), len(relationshipUpdates), batchSize, workers, func(ctx context.Context, no int, start int, end int) error {
```
bryanhuhta, line 178: In theory we could use a cancel context here and have another cli flag to dictate how long a timeout should be. I think that should be a follow up feature though.

internal/grpcutil/batch.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
+)
+
+type batch struct {
+	no    int
+	start int
+	end   int
+}
```
jzelinskie, line 17: I think this is only used in tests now, so we can move this into the test file

jzelinskie (approved):
Technically, this has nothing to do with grpc, so grpcutil probably isn't the best package for it, but it's fine for now.

LGTM

PR #121, opened 2022-05-19
Add DockerHub to goreleaser

josephschorr:


.goreleaser.docker.yml
```diff
@@ -48,6 +48,24 @@ dockers:
     use: "buildx"
     build_flag_templates:
       - "--platform=linux/arm64"
+  - image_templates:
+      - &amd_image "authzed/zed:{{ .Version }}-amd64"
```
ecordell, line 52: isn't this redefining the anchor for `amd_image`? I think this will fail if we try to run it.

in spicedb we did: https://github.com/authzed/spicedb/blob/main/.goreleaser.yml

ecordell (changes requested):

.goreleaser.docker.yml
```diff
@@ -30,18 +30,22 @@ builds:
...
     dockerfile: &dockerfile "Dockerfile.release"
     goos: "linux"
     goarch: "amd64"
     use: "buildx"
     build_flag_templates:
       - "--platform=linux/amd64"
+  # ARM64
   - image_templates:
-      - &arm_image "quay.io/authzed/zed:{{ .Version }}-arm64"
-    ids: ["linux-arm64"]
```
ecordell, line 44: I think you need to leave these id fields

ecordell (changes requested):

.goreleaser.docker.yml
ecordell: there's only 1 `build` in spicedb, but we have to have different ones here since it uses CGO

josephschorr:
Closed since fixed in https://github.com/authzed/zed/pull/143

PR #130, opened 2022-07-19
Add ability to trace a check request

josephschorr:
Provides a trace output when requested, as well as the schema used

Fixes https://github.com/authzed/spicedb/issues/696

Screen Shot 2022-07-19 at 2 48 06 PM
Screen Shot 2022-07-19 at 2 48 09 PM
Screen Shot 2022-07-22 at 6 31 59 PM


josephschorr:
Waiting on changes in API to be merged and pulled into authzed-go

cmd/zed/permission.go
```diff
@@ -24,6 +32,8 @@ func registerPermissionCmd(rootCmd *cobra.Command) {
 	permissionCmd.AddCommand(checkCmd)
 	checkCmd.Flags().Bool("json", false, "output as JSON")
 	checkCmd.Flags().String("revision", "", "optional revision at which to check")
+	checkCmd.Flags().Bool("trace", false, "requests the debug information of the check from SpiceDB and prints out a trace of the requests")
```
jzelinskie, line 35: I'm not convinced that these flag names make sense. `trace` is so overloaded with distributed tracing and logging.

Maybe something like:
`--resolution-path`
`--resolution-schema`

jzelinskie (changes requested):
The display logic should get teased apart from the CLI flag logic and moved to into the `internal/printers` package.

Other than that, this looks good

cmd/zed/permission.go
josephschorr: Those are verbose to type though; I was going for a short, simple name. Maybe... `--debug`?

josephschorr:
> The display logic should get teased apart from the CLI flag logic and moved to into the internal/printers package.

Done

jzelinskie (approved):

Issue #141, opened 2022-08-17
Build distroless debug image variants


Saturas89:
The currently build zed container images based on [distroless](https://github.com/GoogleContainerTools/distroless#debug-images) which ships an extremely minimal environment (it doesn't even include a shell). 

Would be really if you could offer variants of your images using the [distroless debug](https://github.com/GoogleContainerTools/distroless#debug-images) base so that debugging is as simple as changing their image name to authzed/spicedb:$VERSION-debug.

analog to this topic:
https://github.com/authzed/spicedb/issues/749

PR #143, opened 2022-08-17
add debug image variants

vroldanbet:
Fixes https://github.com/authzed/zed/issues/141

I've basically ported changes in https://github.com/authzed/spicedb/pull/750

- I've also updated `Dockerfile` to use distroless.dev/static. The original PR did not update that file.
- I've noticed the `build` section of `.goreleaser.docker.yaml` deviates from the one in SpiceDB. I've left untouched, but wonder if we should align them

### DIFF
--- a/.goreleaser.docker.yml
+++ b/.goreleaser.docker.yml
@@ -30,8 +30,11 @@ builds:
     ldflags: *ldflags
 
 dockers:
+  # AMD64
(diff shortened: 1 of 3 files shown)


jzelinskie:
It deviates a bit because it needs to run on macOS to generate the binary because zed needs cgo for to link against the keychain library.

PR #149, opened 2022-09-13
Add LookupSubjects support to zed

josephschorr:


cmd/zed/permission.go
```diff
@@ -80,6 +104,19 @@ func parseSubject(s string) (namespace, id, relation string, err error) {
...
+	if !strings.Contains(s, "#") {
+		return s, "", nil
+	}
+
+	err = stringz.SplitExact(s, "#", &namespace, &relation)
+	if err != nil {
+		relation = ""
+		err = nil
+	}
+	return
```
jzelinskie, line 117: Because it's only splitting into two things, you're probably better off using [strings.Cut](https://pkg.go.dev/strings#Cut) here. I think by default it's behavior is exactly what we want, so we can even drop the error return variable.

```go
func parseType(s string) (namespace, relation string) {
	namespace, relation, _ = strings.Cut(s, "#")
    return
}
```

jzelinskie (approved):
LGTM, only one minor recommendation for the type splitting code.
This codebase could probably use a nice spring-cleaning for stuff like that, too.

cmd/zed/permission.go
josephschorr: Neat. Changed

jzelinskie (approved):

Issue #161, opened 2022-10-15
Zed Command should allow the use of custom CA Certificates


GauntletWizard:
The zed command currently only uses the system certificate store; However, it's easy enough under the authzed operator to use an internal CA to sign serving certificates. It would be helpful to be able to debug those with the standard zed command.

jzelinskie:
zed will use your system certs -- you could add the CA certificate for your cluster there now.

Do you think it'd make sense to store certificates for each context or have folks specify them each time?

GauntletWizard:
I'm a proponent of per-context; I anticipate you'll want "local", "dev" and "prod" contexts in many cases.

IT Policy explicitly prohibits adding the dev certs to system cert store at most companies I've worked with.

jzelinskie:
Fixed in #183

PR #162, opened 2022-10-15
Allow the use of custom CA Certificates

GauntletWizard:
This allows the use of a custom CA Pool/[TrustBundle](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/3257-trust-anchor-sets) via the `cafile` flag. Starts on #161, but does not support contexts, which would presumably want to embed the ca pool in the context secret.

The decision to use `cafile` for the flag name was based on `curl`, but kubectl calls it `--certificate-authority` and I've seen others.

### DIFF
--- a/cmd/zed/main.go
+++ b/cmd/zed/main.go
@@ -51,16 +51,16 @@ func dialOptsFromFlags(cmd *cobra.Command, token storage.Token) []grpc.DialOptio
 		grpc.WithChainUnaryInterceptor(interceptors...),
 	}
 
-	if cobrautil.MustGetBool(cmd, "insecure") && cobrautil.MustGetString(cmd, "cacert") != "" {
+	if cobrautil.MustGetBool(cmd, "insecure") && cobrautil.MustGetString(cmd, "cafile") != "" {
 		panic("cafile flag cannot be combined with insecure")
 	}
 
 	if cobrautil.MustGetBool(cmd, "insecure") || (token.IsInsecure()) {
 		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
 		opts = append(opts, grpcutil.WithInsecureBearerToken(token.APIToken))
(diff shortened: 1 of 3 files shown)


GauntletWizard:

cmd/zed/main.go
```diff
@@ -51,9 +51,16 @@ func dialOptsFromFlags(cmd *cobra.Command, token storage.Token) []grpc.DialOptio
 		grpc.WithChainUnaryInterceptor(interceptors...),
 	}
 
+	if cobrautil.MustGetBool(cmd, "insecure") && cobrautil.MustGetString(cmd, "cafile") != "" {
+		panic("cafile flag cannot be combined with insecure")
```
jzelinskie, line 55: This should definitely be checked somewhere else and return an error rather than panicking.

internal/storage/secrets.go
```diff
@@ -21,6 +21,7 @@ type Token struct {
 	Endpoint string
 	APIToken string
 	Insecure *bool
+	CAfile   []byte
```
jzelinskie, line 24: Let's just call this field `Certificate`, since it's the actual bytes contents.

jzelinskie (changes requested):
I think `--certificate-path` might be the best name for this flag.

cmd/zed/main.go
GauntletWizard: I don't seen an obvious place to put a "verify flags" precondition; I'm not that familiar with Cobra

jzelinskie: Yeah, that's pretty fair -- I think we should just ignore the CA flag if the insecure flag is set. The insecure flag description is "connect over a plaintext connection", so if you specify it, that's what we should do.

jzelinskie:
Closed in favor of #183.

Let me know if there's anything missing so that we can follow-up.

PR #179, opened 2022-12-30
Rearchitect zed to support a WASM-based mode

josephschorr:
- Moves a bunch of the commands into a shared package for easy use
- Abstracts away writing to the console so it can be overridden
- Abstracts away client creation so it can be overridden
- Moves to cobrautil v2
- Fixes a few small bugs encountered along the way
- Adds a WASM package which allows for zed to be invoked via WASM


internal/console/console.go
```diff
@@ -0,0 +1,17 @@
+package console
+
+import (
+	"fmt"
+)
+
+// Printf defines an (overridable) function for printing to the console.
+var Printf = func(format string, a ...any) {
```
jzelinskie, line 8: There's no way to just override fmt.Printf? This seems like something that should be upstreamed in whatever wasm tooling we're using.

pkg/wasm/client.go
```diff
@@ -0,0 +1,63 @@
+//go:build wasm
```
jzelinskie, line 1: Does this have to be it's own package?

Can we just have this be `pkg/internal/client/client_wasm.go` and `cmd/zed-wasm/main.go`?

pkg/wasm/example/wasm_exec.js
```diff
@@ -0,0 +1,554 @@
+// Copyright 2018 The Go Authors. All rights reserved.
```
jzelinskie, line 1: Can we document where this was sourced from or maybe even have the example load it from there?

pkg/wasm/main.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return zedCommandResult{Error: err.Error()}
+	}
+
+	for _, nsDef := range nsDefs {
+		it, err := reader.QueryRelationships(ctx, datastore.RelationshipsFilter{ResourceType: nsDef.Name})
+		if err != nil {
+			return zedCommandResult{Error: err.Error()}
+		}
+		defer it.Close()
```
jzelinskie, line 175: No need for this defer

pkg/wasm/main.go
```diff
@@ -0,0 +1,212 @@
...
+)
+
+// zedCommandResult is the struct JSON serialized to be returned to the caller.
+type zedCommandResult struct {
+	UpdatedContext string `json:"updated_context"`
+	Output         string `json:"output"`
+	Error          string `json:"error"`
+}
+
+func main() {
```
jzelinskie, line 40: Would it not be possible to have as much shared as possible and then only a little wrapper that sets things up in `cmd/zed/main_wasm.go` and `cmd/zed/main.go`?

jzelinskie (commented):
man, the diff algorithm really fucked this one up.
I've got a few initial comments

.github/workflows/build-test.yaml
```diff
@@ -33,6 +33,21 @@ jobs:
...
+    name: "WASM Tests"
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: "actions/checkout@v3"
+      - uses: "actions/setup-go@v3"
+        with:
+          go-version: "${{ env.GO_VERSION }}"
+          cache: "false" # do not cache to prevent cache poisoning
+      - name: "Install wasmbrowsertest"
+        run: "go install github.com/agnivade/wasmbrowsertest@latest"
```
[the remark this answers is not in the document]
josephschorr: I don't know... I think it might also cause issues with just the different deps pulled in at all

internal/console/console.go
josephschorr: Not as far as I know and when I tried to use a pipe to collect stdout and stderr, it did not work in WASM land

pkg/wasm/client.go
josephschorr: I mean, we could, but this matches how we define things in SpiceDB for WASM as well

pkg/wasm/example/wasm_exec.js
josephschorr: The HTML file actually references its original location. The problem with loading the JS from GitHub is the requirement to, well, load it from GitHub, which is non-ideal, especially if it changes

pkg/wasm/main.go
josephschorr: I always put the `defer it.Close()`, even if I close it after the loop, to *ensure* it will be closed even if the code is changed in the future

pkg/wasm/main.go
josephschorr: This does share as much as I believed I could make shared; the command implementations are all shared, but the actual processing needs to be custom for the JS calling interface

josephschorr:
Updated and rebased

pkg/wasm/example/wasm_exec.js
jzelinskie: If you press `y` on a GitHub page it gives you the absolute URL -- that would prevent from any changes outside of force pushes or repository deletions.

This is just an example, so I'm not super concerned.

pkg/wasm/client.go
josephschorr: Honestly? I prefer having a distinct package. It keeps all the WASM-related stuff in one place and doesn't pollute the other namespaces. As well, this isn't really a "cmd", per se

pkg/wasm/example/wasm_exec.js
josephschorr: Good to know

jzelinskie (approved):

Issue #198, opened 2023-02-16
zed relationship read does not show caveat name and context


vroldanbet:

vroldanbet:
https://github.com/authzed/zed/pull/199

PR #199, opened 2023-02-17
add caveat context to relationship read output

vroldanbet:
Closes https://github.com/authzed/zed/issues/198

the `zed relationship read` command missed caveat context

### DIFF
--- a/internal/commands/relationship.go
+++ b/internal/commands/relationship.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	"github.com/authzed/spicedb/pkg/tuple"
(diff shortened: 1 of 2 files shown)


internal/commands/relationship.go
```diff
@@ -236,15 +237,25 @@ func readRelationships(cmd *cobra.Command, args []string) error {
...
+			relString, err := relationshipToString(msg.Relationship)
+			if err != nil {
+				return err
 			}
+			console.Println(relString)
 		}
 	}
 }
 
+func relationshipToString(rel *v1.Relationship) (string, error) {
```
josephschorr, line 249: Add a unit test for this, since its non-obvious

vroldanbet: addressed

josephschorr (approved):
LGTM

PR #200, opened 2023-02-22
introduce watch command

vroldanbet:
this is useful to test the Watch API and validate
behaviour of a client application that will use the API

### DIFF
--- a/internal/commands/watch.go
+++ b/internal/commands/watch.go
@@ -1,0 +1,88 @@
+package commands
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/authzed/zed/internal/client"
+	"github.com/authzed/zed/internal/console"
+
+	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	"github.com/spf13/cobra"
+)
+
+var (
+	watchObjectTypes []string
+	watchRevision    string
+	watchTimestamps  bool
+)
+
+func RegisterWatchCmd(rootCmd *cobra.Command) *cobra.Command {
+	rootCmd.AddCommand(watchCmd)
+
+	watchCmd.Flags().StringSliceVar(&watchObjectTypes, "object_types", nil, "optional object types to watch updates for")
+	watchCmd.Flags().StringVar(&watchRevision, "revision", "", "optional revision at which to start watching")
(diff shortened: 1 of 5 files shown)


internal/commands/watch.go
```diff
@@ -0,0 +1,69 @@
...
+			console.Println("context done, terminating stream")
+			return nil
+		default:
+			resp, err := watchStream.Recv()
+			if err != nil {
+				return err
+			}
+
+			for _, update := range resp.Updates {
+				console.Printf("%v: %v\n", time.Now(), update)
```
ecordell, line 65: maybe the command should take a `--timestamp` flag for whether you want to see timestamps?

internal/commands/watch.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	req := &v1.WatchRequest{
+		OptionalObjectTypes: watchObjectTypes,
+	}
+	if watchRevision != "" {
+		req.OptionalStartCursor = &v1.ZedToken{Token: watchRevision}
+	}
+
+	watchStream, err := cli.Watch(cmd.Context(), req)
```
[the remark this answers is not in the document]
vroldanbet: ugh, been living a lie all this time 😅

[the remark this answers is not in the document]
vroldanbet: I tested that `Control+C` this thing never ended up in the `ctx.Done()` case of the select clause. 🤷🏻 
Wrapped with cancel

vroldanbet:
@ecordell thanks, addressed your feedback

internal/commands/watch.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	req := &v1.WatchRequest{
+		OptionalObjectTypes: watchObjectTypes,
+	}
+	if watchRevision != "" {
+		req.OptionalStartCursor = &v1.ZedToken{Token: watchRevision}
+	}
+
+	ctx, cancel := context.WithCancel(cmd.Context())
```
ecordell, line 51: don't we still need something like https://github.com/authzed/spicedb/blob/64e3c41567024931d4ae64482299c2ba90452544/pkg/cmd/signals.go#L16 ?

since watch is long running, the only way cancel will get called is if we tie it to the os signals, right?

vroldanbet: yeah, that or if we provide a `--deadline` flag. I can give that a go.

ecordell (approved):

ecordell (approved):

ecordell (approved):

PR #214, opened 2023-03-18
internal/secrets: fix headless file keyring

### LINKED ISSUE #197: Run on Headless Ubuntu / Debian
I just tried to install zed on Ubuntu / Debian headless via homebrew and the .deb package.
Unfortunately upon running `zed context set local localhost:50051 "somerandomkeyhere" --insecure` I get a promt to set a keyring password. Well, I have never set a password and am not using Keyring.

The output I get is:
```
Entering your password a lot? Try setting the environment variable `ZED_KEYRING_PASSWORD`
Enter passphrase to unlock "/home/XYZ/.zed/keyring.jwt":
```

I don't now the password though as I have never set it and entering nothing or a random string all leads to an error.

I would expect the documentation to work which just states for macOS and Linux:
```sh
brew install zed
zed context set local localhost:50051 "somerandomkeyhere" --insecure
zed schema read --insecure
```
https://authzed.com/docs/spicedb/installing#running-via-brew

This might be related to https://github.com/authzed/zed/issues/140.

If keyring is required, this would basically make zed useless in any headless setup I believe.

jzelinskie:
This commit vastly improves the file keyring to
- only prompt once per execution
- specify when creating a password

Fixes #197

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -64,7 +64,8 @@ func DefaultStorage() (storage.ConfigStore, storage.SecretStore) {
 		homedir, _ := homedir.Dir()
 		home = filepath.Join(homedir, ".zed")
 	}
-	return storage.JSONConfigStore{ConfigPath: home}, storage.KeychainSecretStore{ConfigPath: home}
+	return &storage.JSONConfigStore{ConfigPath: home},
+		&storage.KeychainSecretStore{ConfigPath: home}
 }
 
 func certOption(cmd *cobra.Command, token storage.Token) (opt grpc.DialOption, err error) {
--- a/internal/storage/secrets.go
+++ b/internal/storage/secrets.go
@@ -3,7 +3,6 @@ package storage
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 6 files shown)


internal/client/client.go
```diff
@@ -64,7 +64,8 @@ func DefaultStorage() (storage.ConfigStore, storage.SecretStore) {
 		homedir, _ := homedir.Dir()
 		home = filepath.Join(homedir, ".zed")
 	}
-	return storage.JSONConfigStore{ConfigPath: home}, storage.KeychainSecretStore{ConfigPath: home}
+	return &storage.JSONConfigStore{ConfigPath: home},
```
josephschorr, line 67: what changed here besides formatting?

internal/storage/secrets.go
```diff
@@ -114,34 +114,81 @@ func RemoveToken(name string, ss SecretStore) error {
...
 }
 
-const keyringEntryName = "zed secrets"
+var _ SecretStore = (*KeychainSecretStore)(nil)
 
-var _ SecretStore = KeychainSecretStore{}
+const (
+	svcName             = "zed"
+	keyringEntryName    = svcName + " secrets"
+	envRecommendation   = "Entering your password a lot? Try setting the environment variable `ZED_KEYRING_PASSWORD`\n"
```
josephschorr, line 125: to what? we should be explicit here

internal/storage/secrets.go
```diff
@@ -114,34 +114,81 @@ func RemoveToken(name string, ss SecretStore) error {
...
+			// Check if this is the first run where the keyring is created.
+			keyringExists, err := fileExists(filepath.Join(keyringPath, keyringEntryName))
+			if err != nil {
+				return "", err
+			}
+			if !keyringExists {
+				// This is the first run and we're creating a password.
+				prompt = keyringDoesNotExist
+			} else if rand.Int()%10 == 0 { // nolint:gosec
+				// 10% of runs will recommend the environment variable.
```
josephschorr, line 168: why this percentage?

internal/client/client.go
jzelinskie: they're references

internal/storage/secrets.go
jzelinskie: how often do you think we should run it?

josephschorr: Always?

josephschorr (approved):
LGTM

PR #222, opened 2023-04-11
move code in cmd/zed to internal/cmd

vroldanbet:
I was being unable to run tests or run zed from
my IDE of choice, with the following error:

```
go build ./cmd/zed/main.go
# command-line-arguments
cmd/zed/main.go:44:10: undefined: versionCmdFunc
cmd/zed/main.go:55:10: undefined: contextUseCmdFunc cmd/zed/main.go:59:2: undefined: registerContextCmd cmd/zed/main.go:60:2: undefined: registerImportCmd cmd/zed/main.go:61:2: undefined: registerValidateCmd cmd/zed/main.go:69:2: undefined: registerAdditionalSchemaCmds
``

the way the IDE configures the go toolchain would cause some methods to be undefined. While I understand this is specific to one IDE vendor, and that the code would also be effectively unexported in its current state, it did not seem unreasonable to move all this code to the internal package in line with the conventions used in authzed/spicedb.

This commit does not change the code, just moves it to internal, and extracts all code in main.go into its own file with a new exported `Run()` function

### DIFF
--- a/cmd/zed/main.go
+++ b/cmd/zed/main.go
@@ -1,74 +1,7 @@
 package main
 
-import (
-	"os"
-
-	"github.com/jzelinskie/cobrautil/v2"
-	"github.com/jzelinskie/cobrautil/v2/cobrazerolog"
-	"github.com/rs/zerolog"
-	"github.com/spf13/cobra"
-
-	"github.com/authzed/zed/internal/commands"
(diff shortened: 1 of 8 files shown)


josephschorr (approved):
LGTM

PR #223, opened 2023-04-11
replaces cockroachdb treeprinter with github.com/xlab/treeprint

vroldanbet:
despite CRDB's treeprinter converting to Apache 2.0 after the Change Date, the community reported friction on adopting `zed` because automated tools would flag BSL incompatible with Apache 2.0.

This commit replaces cockroachdb treeprinter with
github.com/xlab/treeprint.

It's a mostly 1:1 migration, except due to the design choice to always create an implicit empty root node. I addressed that by removing that implicit root node after the tree is populated, as unfortunately the printer is unable to realize we are starting the print process from an intermediate node. To hide that complexity I wrapped it in our own type.

### DIFF
--- a/internal/commands/permission.go
+++ b/internal/commands/permission.go
@@ -10,7 +10,6 @@ import (
 	"github.com/authzed/authzed-go/pkg/requestmeta"
 	"github.com/authzed/authzed-go/pkg/responsemeta"
 	v1 "github.com/authzed/authzed-go/proto/authzed/api/v1"
-	"github.com/cockroachdb/cockroach/pkg/util/treeprinter"
 	"github.com/jzelinskie/cobrautil/v2"
 	"github.com/jzelinskie/stringz"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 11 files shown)


internal/commands/permission.go
```diff
@@ -281,8 +281,12 @@ func expandCmdFunc(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	tp := treeprinter.New()
+	tp := treeprint.New()
 	printers.TreeNodeTree(tp, resp.TreeRoot)
+	// remove implicit root node in `treeprint.New` call
```
josephschorr, line 286: newlines before comments and can you expand on this comment a bit?

internal/commands/permission.go
```diff
@@ -471,9 +475,12 @@ func displayDebugInformationIfRequested(cmd *cobra.Command, trailerMD metadata.M
 		}
 
 		if cobrautil.MustGetBool(cmd, "explain") {
-			tp := treeprinter.New()
+			tp := treeprint.New()
```
josephschorr, line 478: maybe we should do this in a function, since we're repeating it here?

vroldanbet:
@josephschorr reimplemented with a wrapper that hides the complexity of the design choices of `treeprint` module

jzelinskie (approved):

PR #251, opened 2023-06-21
Backup zedtoken

### LINKED ISSUE #249: Store backup snapshot timestamp in the backupfile
We currently export a backup (schema and relationships) from a single backup snapshot timestamp. We should encode that timestamp in the backupfile to enable snapshot + delta synchronization between permissions systems.

jakedt:
Fixes #249

### DIFF
--- a/internal/cmd/restore.go
+++ b/internal/cmd/restore.go
@@ -22,6 +22,7 @@ func registerRestoreCmd(rootCmd *cobra.Command) {
 	rootCmd.AddCommand(restoreCmd)
 	restoreCmd.Flags().Int("batch-size", 1_000, "restore relationship write batch size")
 	restoreCmd.Flags().Int("batches-per-transaction", 10, "number of batches per transaction")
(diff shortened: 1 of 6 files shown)


internal/cmd/restore.go
```diff
@@ -22,6 +22,7 @@ func registerRestoreCmd(rootCmd *cobra.Command) {
 	rootCmd.AddCommand(restoreCmd)
 	restoreCmd.Flags().Int("batch-size", 1_000, "restore relationship write batch size")
 	restoreCmd.Flags().Int("batches-per-transaction", 10, "number of batches per transaction")
+	restoreCmd.Flags().Bool("print-zedtoken-only", false, "just print the zedtoken and stop")
```
jzelinskie, line 25: Maybe this should be another command entirely?

jzelinskie (dismissed):
LGTM only an open ended UX question

internal/cmd/restore.go
jakedt: I mean, I considered it, and thought that this would be more discoverable.

jzelinskie (approved):

PR #263, opened 2023-07-08
Allow calls to zed relationship {create, touch, delete} to read from stdin

epaulson10:
One common thing I find myself doing is:

```shell
zed relationship read someResource > relationships.txt
vim relationships.txt # perform some bulk edit
cat relationships.txt | xargs -n 3 zed relationship touch
```
Using xargs here works fine, but it requires creating a brand new process and gRPC connection for each relationship, which is quite slow.

Also, with the addition of caveats, you can no longer reliably perform this, as the caveat output of `zed relationship read` would break the above example, since caveat context can only be specified via `--caveat`.

This PR is how I have addressed the above in my own fork of this tool, and I would like to offer it back upstream. This PR does two things:

1. Changes the parsing logic for the arguments of `zed relationship {create, touch, delete}` to use the tuple package from SpiceDB, which can handle specifying the caveat context at the end of the subject. This ensures that anything output by `zed relationship read` can be passed directly back into `zed relationship {create, touch, delete}`.
    - I left the --caveat flag in, and returned an error if both a relationship specifies a caveat, and the flag is present.
2. Allows the update commands to read a stream of newline-separated relationships from stdin, with configurable batching to further improve write performance within scripts.


epaulson10:

epaulson10:
recheck

vroldanbet (commented):
@epaulson10 thanks for the contribution! I see the PR adds a batch-size flag. Did you use these changes for bulk loading into SpiceDB? Out of curiosity have you looked into the new bulk load API? Mind sharing what was your use-case?

epaulson10:
> Did you use these changes for bulk loading into SpiceDB?

Not quite. This change is more to make zed more useful in shell scripting scenarios. For example, let's say you have a `group` and an `organization` type defined, and you want to run a script that grabs all the `member` relations from an `organization` and assigns them to a group. The only way to do this using zed alone is to write each individual relationship using `zed relationship {create,touch}`, which has a pretty high overhead, as you are spinning up a whole process and gRPC connection each time. For small amounts of relationships, it is fine, but if you have 1000s of relationships, the script is going to take quite a while to run. (I know you could probably do some schema-wrangling to directly assign the membership relation of the `organization` to the `group`, but I'm just trying to come up with an example).

>  Out of curiosity have you looked into the new bulk load API? 

I have looked at it. Given it's not available in zed yet, it's not of immediate use to me. I've been using zed in scripts to manage permissions until more robust tooling is built around the APIs on my end. Also, I'm a little cautious about using experimental APIs.

If you all would rather put a new command into zed that does something similar to this PR, but using the `BulkImportRelationships` API, I would be happy to rewrite this to do so. I'm not sure how well that would fit though, given the ask to try to get the data into a certain sort-order. Also, I personally think it is useful to expose writing multiple relationships using the `WriteRelationships` API, since that API does support it. Another idea I thought about was to allow the user to supply unbounded arguments in multiples of 3, to allow writing multiple relationships without needing to pipe data through `stdin` (though reading from stdin is very cool and Unixy).

vroldanbet (commented):
>>Did you use these changes for bulk loading into SpiceDB?

>Not quite. This change is more to make zed more useful in shell scripting scenarios. For example, let's say you have a group and an organization type defined, and you want to run a script that grabs all the member relations from an organization and assigns them to a group. The only way to do this using zed alone is to write each individual relationship using zed relationship {create,touch}, which has a pretty high overhead, as you are spinning up a whole process and gRPC connection each time. For small amounts of relationships, it is fine, but if you have 1000s of relationships, the script is going to take quite a while to run. (I know you could probably do some schema-wrangling to directly assign the membership relation of the organization to the group, but I'm just trying to come up with an example).

Right. To be clear I think your change is very useful and I'd like to see it merged, but wanted to understand the underlying use-case, particularly because if it's about backfilling a SpiceDB instance then loading large amounts of relationships through `WriteRelationships` will also suffer from overhead even if you send batches of them. This is why we created the bulk APIs. 

>>Out of curiosity have you looked into the new bulk load API?
>I have looked at it. Given it's not available in zed yet, it's not of immediate use to me.

I think that's probably an oversight, as we were waiting for the corresponding SpiceDB version with Experimental API support released. I've opened https://github.com/authzed/zed/releases/tag/untagged-a944e43474738ef4898f
and will be looking into releasing it today.

>I've been using zed in scripts to manage permissions until more robust tooling is built around the APIs on my end. Also, I'm a little cautious about using experimental APIs.

Totally fair! As you say, I can also see zed very useful for those scenarios where you can throw around some shell scripting until you build the application in charge of this (if at all!). Anything that lowers the barrier of entry is a good thing.
 
>If you all would rather put a new command into zed that does something similar to this PR, but using the BulkImportRelationships API, I would be happy to rewrite this to do so. I'm not sure how well that would fit though, given the ask to try to get the data into a certain sort-order. Also, I personally think it is useful to expose writing multiple relationships using the WriteRelationships API, since that API does support it.

I don't think they are mutually exclusive. The backup and restore commands will be released soon, but this PR is IMHO still useful.

>Another idea I thought about was to allow the user to supply unbounded arguments in multiples of 3, to allow writing multiple relationships without needing to pipe data through stdin (though reading from stdin is very cool and Unixy).

We definitely want the tool to follow Unix program principles. If the community thinks unbounded arguments also have their place, we can look into it.

Thanks for the thoughtful contribution! We will be looking into it soon

vroldanbet:
@epaulson10 I took the liberty to push some changes as part of my review, hope that's ok, happy to discuss the changes 🙏🏻 . I only have pending to review the `writeRelationshipCmdFunc` method.

epaulson10:
@vroldanbet , those changes look good to me. I initially wanted to do something similar with `IndexFunc`, but I was concerned about potential duplicate spaces between the arguments. It turns out I forgot that `TrimSpace` existed :sweat_smile:

vroldanbet:
@epaulson10 rebased the PR on top of `main` and added https://github.com/authzed/zed/pull/263/commits/0661a39eb14c2610f5dd75b8cc3cd10c45896c05 which refactors `writeRelationshipCmdFunc` and adds a bunch of tests. Let me know what you think!

internal/commands/relationship.go
```diff
@@ -250,100 +279,197 @@ func readRelationships(cmd *cobra.Command, args []string) error {
...
+	}
+	return argsToRelationship(args)
+}
+
+func writeRelationshipCmdFunc(operation v1.RelationshipUpdate_Operation, clientFunc func(command *cobra.Command) (client.Client, error), input *os.File) func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		// getNextRelationship is a function that will fetch the next relationship to write.
+		// Either it will read a stream of relationships from stdin, or it will simply return
+		// the command line arguments once. When there are no more relationships, it will return
+		// nil for the first parameter.
```
jzelinskie, line 405: Why not make this a bit more explicit in the type system and create a place where this comment can live:

```go
// RelationshipParser is a closure that can produce relationships.
// When there are no more relationships, it will return ErrExhaustedRelationships.
type RelationshipParser func() (*v1.Relationship, error)

// ErrExhaustedRelationships signals that the last producible value of a RelationshipParser
// has already been consumed.
// Functions should return this error to signal a graceful end of input.
var ErrExhaustedRelationships = errors.New("exhausted all relationships")
```

```go
func FileRelationshipParser(f *os.File) RelationshipParser {
	scanner := bufio.NewScanner(file)
	return func() (*v1.Relationship, error) {
		if scanner.Scan() {
			res, rel, subj, err := parseRelationshipLine(scanner.Text()) // Notice this signature change
			if err != nil {
				return nil, err
			}
			return tupleToRel(res, rel, subj) // This does your sprintf + string parsing
		}
		return nil, scanner.Err()
	}
}
```

If you refactor the args func like this too, the call site gets cleaned up a bit:

```go
	return func(cmd *cobra.Command, args []string) error {
		parse := SliceRelationshipParser(args)
		if !isFileTerminal(input) {
			// If it's not a terminal, assume the program is being piped into.
			parse = FileRelationshipParser(input)
		}
		...
		for {
			rel, err := parse()
			if errors.Is(err, ErrRelationshipsExhausted) {
				break
			} else if err != nil {
				return err
			}
			...
		}

```

internal/commands/relationship.go
```diff
@@ -1,36 +1,45 @@
...
 	"github.com/jzelinskie/stringz"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
-
-	"github.com/authzed/zed/internal/client"
-	"github.com/authzed/zed/internal/console"
+	"golang.org/x/term"
 )
 
+var isTerminal = term.IsTerminal
```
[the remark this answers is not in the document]
vroldanbet: `term.IsTerminal` was already used in other parts of the codebase., so I'm not sure if that was already exercised in wasm. I can add a write relationships call in the wasm test.

internal/commands/relationship.go
```diff
@@ -87,8 +96,28 @@ var bulkDeleteCmd = &cobra.Command{
 	RunE:  bulkDeleteRelationships,
 }
 
+func writeRelationshipsArgsWithStdin(cmd *cobra.Command, args []string) error {
+	return writeRelationshipsArgs(cmd, args, os.Stdin)
+}
+
+func writeRelationshipsArgs(cmd *cobra.Command, args []string, file *os.File) error {
```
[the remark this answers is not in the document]
vroldanbet: this was needed before when we didn't have the mock in place. The mocking was a last-minute change when Actions started to fail, so this is not required anymore, I will address it.

internal/commands/relationship.go
```diff
@@ -55,22 +64,22 @@ var relationshipCmd = &cobra.Command{
...
+	Args:  writeRelationshipsArgsWithStdin,
+	RunE:  writeRelationshipCmdFunc(v1.RelationshipUpdate_OPERATION_TOUCH, client.NewClient, os.Stdin),
 }
 
 var deleteCmd = &cobra.Command{
 	Use:   "delete <resource:id> <relation> <subject:id>",
 	Short: "delete a Relationship",
-	Args:  cobra.ExactArgs(3),
-	RunE:  writeRelationshipCmdFunc(v1.RelationshipUpdate_OPERATION_DELETE),
+	Args:  writeRelationshipsArgsWithStdin,
```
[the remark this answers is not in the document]
vroldanbet: I removed it anyway

internal/commands/relationship.go
vroldanbet: I was a bit lazy to refactor this a bit further, thanks for doing it!

internal/commands/relationship.go
```diff
@@ -250,100 +270,203 @@ func readRelationships(cmd *cobra.Command, args []string) error {
 	}
 }
 
+func argsToRelationship(args []string) (*v1.Relationship, error) {
```
jzelinskie, line 273: It looks like we spotted using this in favor of tupleToRel, but this does additional validation because `tuple.ParseRel` checks for valid/invalid characters. We should unify the usage and make sure we're using the `tuple.ParseRel`

vroldanbet: good catch!

PR #264, opened 2023-07-08
Fix bug when writing schema supplied via stdin

epaulson10:
The check to make sure that stdin is not a terminal was actually checking stdout, meaning that you cannot actually send a schema via stdin. This commit fixes this check.

I noticed this when working on #263 .


vroldanbet (approved):
Good catch, LGTM! Thanks for your contribution! 🌟

PR #269, opened 2023-07-24
fixes when STDIN input and CLI args are provided

vroldanbet:
a potentially common scenario is zed is being invoked from a shell script that is having its input piped, but zed itself is being provided with arguments as the result of parsing that input in the script. In such case arguments provided should take precedence to the STDIN, instead of erroring out


vroldanbet:
FYI @epaulson10

PR #283, opened 2023-08-09
Add support for outputting dispatch counts on streaming APIs

josephschorr:


internal/client/client.go
```diff
@@ -93,7 +91,10 @@ func DialOptsFromFlags(cmd *cobra.Command, token storage.Token) ([]grpc.DialOpti
 		interceptors = append(interceptors, zgrpcutil.CheckServerVersion)
 	}
 
-	opts := []grpc.DialOption{grpc.WithChainUnaryInterceptor(interceptors...)}
+	opts := []grpc.DialOption{
+		grpc.WithChainUnaryInterceptor(interceptors...),
+		grpc.WithChainStreamInterceptor(zgrpcutil.StreamLogDispatchTrailers),
```
ecordell, line 96: shouldn't we implement a streaming version of the version check interceptor too, and call it here the same way as we do for unary?

josephschorr: Mind if I do that as a followup?

ecordell (approved):

Issue #284, opened 2023-08-17
zed backup to stdout or existing file


FrankFoerster24:
Hi,

I would like to send the backup to some subsequent processing, eg. compression, encryption, and other voodoo before hitting the final backup target.  I would very much like if I could make it write to a previously created named pipe which I could read from with other programs.  Maybe a `--force` overriding the [check](https://github.com/authzed/zed/blob/main/internal/cmd/backup.go#L37-L39) would already do the trick.

Alternatively, sending the output to `stdout` to pipe it to other tools could also work.

Kind regards,
 Frank.

FrankFoerster24:
Hi,

thank you very much for the swift implementation.  However, for completeness I'd like to add that I still have the use case to send the backup data to an existing file (which is AFAICS not possible right now).

This is my use case: I'd like to run backups using the provided docker images which don't have much other utilities, not even a shell to `>` redirect the stream on `stdout` AFAIK.  I'm not keen to run my own images.  The pattern I usually would use is that I send the data to a named pipe (created by `mkfifo` on a shared volume in an init container), have the `zed` container write to it and another container in the same pod read from it for further processing (compression, encryption, etc.).  With `zed` refusing to write to the existing named pipe I cannot do that.

I'll find workarounds, but the described pattern would be fit most perfectly for me.  Let me know how to proceed (reopen this issue, create a new one, go away and leave you alone... ;-)).

Thanks again and kind regards,
 Frank.

miparnisari:
It seems that we broke support for this in v0.28.0. Reopening.


```
[12:21:05] ~/Documents/GitHub/zed (main) $ zed backup create -

// ommitted for brevity

12:21PM ERR terminated with errors error="failed to store cursor in lockfile: cannot create temp file: open /dev/stdout.lock2654921088: operation not permitted\nsync /dev/stdout: inappropriate ioctl for device"
```

@ivanauth i left a comment to that effect in https://github.com/authzed/zed/pull/688#discussion_r3183888107

ivanauth:
Fix is in #691 - drops the lockfile/Sync/Close path when the destination is -. Adds a TestFileEncoderStreamingToStdout covering the regression.

PR #690 separately adds tests for printBackupFileLocation per the review on #688.

PR #285, opened 2023-08-17
Support stdout / stdin for backup and restore

bison:
This adds support for passing a single dash character as the filename
to the backup command, resulting in it using stdout.  It also changes
the restore command to default to reading from stdin if no filename is
given.

Fixes: #284

### DIFF
--- a/internal/cmd/backup.go
+++ b/internal/cmd/backup.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


internal/cmd/restore.go
```diff
@@ -32,27 +32,39 @@ var restoreCmd = &cobra.Command{
 	RunE:  restoreCmdFunc,
 }
 
-func restoreCmdFunc(cmd *cobra.Command, args []string) error {
-	filename := args[0]
+func openRestoreFile(filename string) (*os.File, int64, error) {
+	if filename == "-" {
```
bison, line 36: Instead of using `-` here, should `stdin` just be the default if no filename is given?

bison: Done.

internal/cmd/backup.go
```diff
@@ -116,8 +127,10 @@ func backupCmdFunc(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("error closing backup encoder: %w", err)
 	}
 
-	if err := f.Sync(); err != nil {
-		return fmt.Errorf("error syncing backup file: %w", err)
+	if f != os.Stdout {
```
jzelinskie, line 130: Shouldn't sync still work on os.Stdout? It's still a *File.

bison: You can call it, but it will return an error.

bison: If we can tell it's an `inappropriate ioctl for device` error, we could just specifically ignore that error instead.

jzelinskie: Screenshot 2023-08-23 at 1 12 06 PM
Looks like there are multiple types of errors. I think this code is reasonable to avoid trying to catch a million error cases (especially if they differ cross platform).

jzelinskie (approved):

PR #288, opened 2023-08-22
update read relationships to use pagination

vroldanbet:
it's backward compatible with SpiceDB instances
that do not support pagination, and does this by
not asking for another page if the page size was
bigger than the limit.

Tested with SpiceDB 1.24.0 and 1.19.0

Sets the default page size to 100

### DIFF
--- a/internal/commands/relationship.go
+++ b/internal/commands/relationship.go
@@ -43,6 +43,7 @@ func RegisterRelationshipCmd(rootCmd *cobra.Command) *cobra.Command {
 	readCmd.Flags().String("revision", "", "optional revision at which to check")
 	_ = readCmd.Flags().MarkHidden("revision")
 	readCmd.Flags().String("subject-filter", "", "optional subject filter")
+	readCmd.Flags().Uint32("page-limit", 100, "limit of relations returned per page")
 	registerConsistencyFlags(readCmd.Flags())
 
 	relationshipCmd.AddCommand(bulkDeleteCmd)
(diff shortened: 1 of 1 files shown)


internal/commands/relationship.go
```diff
@@ -43,6 +43,7 @@ func RegisterRelationshipCmd(rootCmd *cobra.Command) *cobra.Command {
 	readCmd.Flags().String("revision", "", "optional revision at which to check")
 	_ = readCmd.Flags().MarkHidden("revision")
 	readCmd.Flags().String("subject-filter", "", "optional subject filter")
+	readCmd.Flags().Uint32("page-limit", 100, "limit of relationships to return")
```
jzelinskie, line 46: `limit of relations returned per page`

internal/commands/relationship.go
```diff
@@ -209,60 +210,79 @@ func buildReadRequest(cmd *cobra.Command, args []string) (*v1.ReadRelationshipsR
...
+		log.Trace().Interface("request", request).Msg("reading relationships page")
+		readRelClient, err := spicedbClient.ReadRelationships(cmd.Context(), request)
 		if err != nil {
 			return err
 		}
 
-		if cobrautil.MustGetBool(cmd, "json") {
-			prettyProto, err := PrettyProto(msg)
-			if err != nil {
+		relCount := uint32(0)
```
jzelinskie, line 246: var relCount uint32

internal/commands/relationship.go
```diff
@@ -209,60 +210,79 @@ func buildReadRequest(cmd *cobra.Command, args []string) (*v1.ReadRelationshipsR
...
-			return err
-		}
-
-		msg, err := resp.Recv()
-		if errors.Is(err, io.EOF) {
-			return nil
-		}
-
+		request.OptionalCursor = lastCursor
+		log.Trace().Interface("request", request).Msg("reading relationships page")
```
jzelinskie, line 240: If we're tracing maybe we should also log the cursor itself

internal/commands/relationship.go
```diff
@@ -209,60 +210,79 @@ func buildReadRequest(cmd *cobra.Command, args []string) (*v1.ReadRelationshipsR
...
+				}
+
+				console.Println(string(prettyProto))
+			} else {
+				relString, err := relationshipToString(msg.Relationship)
+				if err != nil {
+					return err
+				}
+				console.Println(relString)
+			}
```
jzelinskie, line 276: This seems like a good thing to pull out into it's own function.
I was a bit confused by the `relCount++` then a conditional, but I wouldn't have been if i saw `relCount++; printRel(...)`

jzelinskie (changes requested):

jzelinskie (approved):

Issue #303, opened 2023-10-02
The context token doesn't appear to be using the stored NoVerifyCA and CACert


prufrock:
The context token supports NoVerifyCA and CACert. When it restores the values from a keychain it isn't mapping the values from the token read to the token returned. This prevents zed client users from benefiting from stored --no-verify-ca and --certificate-path values.

It looks like a pretty quick fix, so I should have a PR attached shortly.

PR #304, opened 2023-10-02
Add missing context token attributes

prufrock:
The context token supports NoVerifyCA and CACert. When it restored the values from a keychain it wasn't mapping the values from the token read to the token returned. This prevented zed client users from benefiting from stored --no-verify-ca and --certificate-path values.

With a simple addition the values are passed from context and they no longer need to be manually set on the command line.


prufrock:
Fixes #303

prufrock:

ecordell (approved):
LGTM

PR #310, opened 2023-11-23
restore: report error if one of the batches fail

vroldanbet:
If one of the batches sent fails, zed will hide the underlying error and just report `EOF`.

The gRPC backend will report EOF as part of an error, which will have to be retrieved via `CloseAndRecv()`.

A common error that surfaced as EOF was duplicate relationships after reimporting the same backup file.

### DIFF
--- a/internal/cmd/restore.go
+++ b/internal/cmd/restore.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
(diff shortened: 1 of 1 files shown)


vroldanbet:
merged via https://github.com/authzed/zed/pull/313

PR #313, opened 2023-12-08
Backups: subcommands, prefix-filtering, parsing

jzelinskie:
This change 

- reorganizes backup functionality to match the format of the rest of the zed CLI's `zed $noun $verb` structure
- adds a `--prefix-filter` flag to commands to filter schemas and relationships from backup/restore/parsing
- adds `zed backup parse-schema`, `zed backup parse-revision`, `zed backup parse-relationships` commands to inspect backup files


internal/cmd/backup.go
```diff
@@ -48,7 +71,53 @@ func createBackupFile(filename string) (*os.File, error) {
 	return f, nil
 }
 
-func backupCmdFunc(cmd *cobra.Command, args []string) error {
+var (
+	missingAllowedTypes = regexp.MustCompile(`(\s*)(relation)(.+)(\/\* missing allowed types \*\/)(.*)`)
```
jakedt, line 75: What stops this from getting greedy and gobbling up more than we anticipated?

internal/cmd/backup.go
```diff
@@ -48,7 +71,53 @@ func createBackupFile(filename string) (*os.File, error) {
...
+	filteredSchema, _, err = generator.GenerateSchema(prefixedDefs)
+	if err != nil {
+		return "", fmt.Errorf("error generating filtered schema: %w", err)
+	}
+	return
+}
+
+func hasRelPrefix(rel *v1.Relationship, prefix string) bool {
+	// Skip any relationships without the prefix on either side.
+	return strings.HasPrefix(rel.Resource.ObjectType, prefix) ||
```
jakedt, line 116: I wonder if we want to do this for either side, or if we only want to include relationships that originate on the prefix side...

internal/cmd/backup.go
```diff
@@ -78,7 +147,17 @@ func backupCmdFunc(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("`backup` is not supported on this version of SpiceDB")
 	}
 
-	encoder, err := backupformat.NewEncoder(relWriter, schemaResp.SchemaText, schemaResp.ReadAt)
+	// Skip any definitions without the provided prefix
+	schema := schemaResp.SchemaText
+	prefixFilter := cobrautil.MustGetString(cmd, "prefix-filter")
+	if prefixFilter != "" {
+		schema, err = filterSchemaDefs(schema, prefixFilter)
```
jakedt, line 154: There is the opportunity here to provide a prefix of `fire`, expecting to get only `fire/` but also getting `firetruck/` on accident. Do we want to add any logic around the separators?

jakedt (dismissed):
Just some questions.

internal/cmd/backup.go
jzelinskie: Not much other than the low odds of someone having exactly this match. I could hide this behind a flag and have it default off.

internal/cmd/backup.go
jzelinskie: I can change it so that it's everything to the left of the final `/` just to be 1000% sure

jzelinskie: ✅

internal/cmd/backup.go
jzelinskie: ✅ put this behind a flag so only folks that need it will use it

internal/cmd/backup.go
jzelinskie: Either way you need the definition, right? I wonder if there's something from SpiceDB's reachability analysis that could be used here for maximum correctness.

jzelinskie: I used the typesystem to guarantee that our generated schema is comprehensive, this means that if there was any non-prefixed dependencies, we would throw a failure.

We can circle back later and support for discovering the dependencies.

internal/commands/relationship.go
```diff
@@ -94,11 +94,14 @@ var bulkDeleteCmd = &cobra.Command{
 	RunE:  bulkDeleteRelationships,
 }
 
-func writeRelationshipsFromArgsOrStdin(cmd *cobra.Command, args []string) error {
-	if ok := isArgsViaFile(os.Stdin) && len(args) == 0; ok {
-		return nil
+func StdinOrExactArgs(n int) cobra.PositionalArgs {
```
vroldanbet, line 97: nice

internal/cmd/backup.go
```diff
@@ -111,39 +290,307 @@ func backupCmdFunc(cmd *cobra.Command, args []string) error {
...
+					return fmt.Errorf("error storing relationship: %w", err)
+				}
+				relsEncoded++
 
-			if stored%100_000 == 0 && !hasProgressbar {
-				log.Trace().Uint("relationships", stored).Msg("relationships stored")
+				if relsEncoded%100_000 == 0 && !isatty.IsTerminal(os.Stderr.Fd()) {
+					log.Trace().
+						Uint("encoded", relsEncoded).
+						Uint("processed", relsProcessed).
```
vroldanbet, line 302: nit: a percentage value is easier to parse when someone is looking at the logs

vroldanbet (commented):
just one question, otherwise LGTM

internal/commands/relationship.go
jzelinskie: yeah i should probably throw this into cobrautil, haha

internal/cmd/backup.go
jzelinskie: we don't know the total number of relationships in the database

vroldanbet (approved):

internal/cmd/backup.go
```diff
@@ -48,6 +54,52 @@ func createBackupFile(filename string) (*os.File, error) {
...
+	shortRelations      = regexp.MustCompile(`(\s*)relation [a-z][a-z0-9_]:(.+)`)
+)
+
+func filterSchemaDefs(schema, prefix string) (filteredSchema string, err error) {
+	// Remove any invalid relations generated from old, backwards-incompat
+	// Serverless permission systems.
+	schema = string(missingAllowedTypes.ReplaceAll([]byte(schema), []byte("\n/* deleted missing allowed type error */")))
+	schema = string(shortRelations.ReplaceAll([]byte(schema), []byte("\n/* deleted short relation name */")))
+
+	compiledSchema, err := compiler.Compile(compiler.InputSchema{Source: "schema", SchemaString: schema}, compiler.SkipValidation())
```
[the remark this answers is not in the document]
vroldanbet: let's not block on this since at least is a safe strategy. I think the scenario I reproduced with a test was precisely what the regexes above fix, so it may be difficult to actually make it err

PR #316, opened 2023-12-18
refactor backup restore to handle serialization errors and conflicts

vroldanbet:
Related to https://github.com/authzed/spicedb/issues/1544
Related to https://github.com/authzed/spicedb/issues/1542

This commit refactors the logic on the backup command into a restorer type. The fundamental differences with the previous logic are:
- batch conflicts are detected, and provides to handling strategies: fail, skip or touch when using touch, they will also be retried if needed
- serialization errors are detected, and optionally can be retried using WithRelationships with TOUCH semantics
- retry strategy uses a backoff

any error on bulk import will cause the stream to close, so attempts to recover from that by retrying means retrying with another bulk import stream or using WriteRelationships. The latter was used because it normalizes retries as BulkImport does not support TOUCH semantics.

## use

New flags are added to the `zed backup restore` command:
- `--conflict-strategy` defaults to `fail`, which means `zed` will fail if a conflicting relationship is found. This is kept around to remain backward compatible with the original behavior.
  - `skip` would skip importing any batch that contains a conflicting relationship. Please note this may skip relationships that are part of the batch but not stored in SpiceDB
  - `touch` will issue a `WriteRelationships` call with each batch
`--disable-retries` is added to perform retries, which is the original behavior. Once this PR lands, retries will be enabled by default. The maximum number of retries is 10 and has an exponential backoff. Both are non-configurable.
`--request-timeout` is added to add a timeout to `WriteRelationship` calls used on conflicts. `BulkImport` API calls are not subject to the timeout.


internal/cmd/backup.go
```diff
@@ -87,6 +87,9 @@ func registerBackupCmd(rootCmd *cobra.Command) {
 	backupCmd.AddCommand(backupRestoreCmd)
 	backupRestoreCmd.Flags().Int("batch-size", 1_000, "restore relationship write batch size")
 	backupRestoreCmd.Flags().Int64("batches-per-transaction", 10, "number of batches per transaction")
+	backupRestoreCmd.Flags().Bool("skip-conflicts", false, "skips any batch found to be conflicting")
```
josephschorr, line 90: are `skip-conflicts` and `touch-conflicts` exclusive? If so, maybe we have a combined flag with an enum option?

internal/cmd/backup.go
```diff
@@ -377,122 +380,33 @@ func backupRestoreCmdFunc(cmd *cobra.Command, args []string) error {
...
-	// Write the last batch
-	if err := relationshipWriter.Send(&v1.BulkImportRelationshipsRequest{
-		Relationships: batch,
-	}); err != nil {
-		return fmt.Errorf("error sending last batch to server: %w", err)
+	skipConflicts := cobrautil.MustGetBool(cmd, "skip-conflicts")
+	touchConflicts := cobrautil.MustGetBool(cmd, "touch-conflicts")
+	disableRetries := cobrautil.MustGetBool(cmd, "disable-retries")
+	if touchConflicts && skipConflicts {
+		return errors.New("cannot use --skip-conflicts and --touch-conflicts together")
```
josephschorr, line 402: yeah, see above

internal/cmd/restorer.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/samber/lo"
+	"github.com/schollz/progressbar/v3"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/authzed/zed/internal/client"
+	"github.com/authzed/zed/pkg/backupformat"
+)
+
+// Fallback for CRDB datastore on SpiceDB < 1.29.0 not returning proper gRPC codes
```
josephschorr, line 23: what about the other datastore kinds?

internal/cmd/restorer.go
```diff
@@ -0,0 +1,350 @@
...
+	txConflictCodes     = []string{"SQLSTATE 23505"}
+	retryableErrorCodes = []string{"retryable error"}
+)
+
+type restorer struct {
+	decoder               *backupformat.Decoder
+	client                client.Client
+	prefixFilter          string
+	batchSize             int
+	batchesPerTransaction int64
```
josephschorr, line 34: uint64?

internal/cmd/restorer.go
```diff
@@ -0,0 +1,350 @@
...
+		updates := lo.Map[*v1.Relationship, *v1.RelationshipUpdate](batch, func(item *v1.Relationship, index int) *v1.RelationshipUpdate {
+			return &v1.RelationshipUpdate{
+				Relationship: item,
+				Operation:    v1.RelationshipUpdate_OPERATION_TOUCH,
+			}
+		})
+
+		for {
+			// throttle the writes so we don't overwhelm the server
+			time.Sleep(backoffInterval.NextBackOff())
```
josephschorr, line 281: add a log here, perhaps changing the progress bar message temporarily?

internal/grpcutil/grpcutil.go
```diff
@@ -40,51 +43,53 @@ func CheckServerVersion(
...
-		log.Debug().Msg("error reading server version response header; it may be disabled on the server")
-	} else if len(version) == 1 {
-		currentVersion := version[0]
-
-		// If there is a build on the version, then do not compare.
-		if semver.Build(currentVersion) != "" {
-			log.Debug().Str("this-version", currentVersion).Msg("received build version of SpiceDB")
-			return nil
-		}
+	once.Do(func() {
```
josephschorr, line 46: why this change?

vroldanbet: yeah this wasn't obvious: this is run as a middleware on each API call. The `backup restore` command will perform many API calls, so this ended up spamming those messages. It's only really necessary "once per zed session".

internal/cmd/backup.go
vroldanbet: yeah Jimmy shared recently https://github.com/thediveo/enumflag - I'll look into it

internal/cmd/restorer.go
vroldanbet: I didn't test all of them. I'll look into it.

internal/cmd/restorer.go
vroldanbet: yeah better to change the progress bar - logs mixed with the progress bar do not render well.

internal/cmd/backup.go
jzelinskie: yeah it's a bit of a shame because i think it only works with flags registered with `Var()`.

either way we should make this something like `--on-conflict error/skip/touch`

josephschorr:
Reference: https://github.com/authzed/spicedb/issues/1544

internal/cmd/restorer.go
vroldanbet: I tested all datastores and added codes to handle when it was missing

internal/cmd/backup.go
vroldanbet: Addressed. Ended up creating a `Must*` style of helper we could add to `cobrautil`

internal/cmd/restorer.go
vroldanbet: addressed

internal/cmd/cmd.go
```diff
@@ -27,7 +32,14 @@ func Run() {
...
 			SyncFlagsCmdFunc,
 		),
+		SilenceErrors: true,
+		SilenceUsage:  true,
 	}
+	rootCmd.SetFlagErrorFunc(func(cmd *cobra.Command, err error) error {
+		cmd.Println(err)
+		cmd.Println(cmd.UsageString())
+		return errParsing
+	})
```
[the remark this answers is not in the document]
vroldanbet: We use the same pattern in SpiceDB. The idea is that any failure does not cause `zed` to output the help. Before if, say, a bulk import failed, it would output the error, and then the usage.

Now the help blob is only output if there is a parsing error (i.e. incorrect command/flags usage)

internal/cmd/backup.go
```diff
@@ -377,122 +393,39 @@ func backupRestoreCmdFunc(cmd *cobra.Command, args []string) error {
...
-	return nil
-}
-
-func perSec(i uint64, d time.Duration) uint64 {
-	secs := uint64(d.Seconds())
-	if secs == 0 {
-		return i
-	}
-	return i / secs
+	panic(fmt.Sprintf("unexpected flag value %s: should be one of %v", value, maps.Keys(mapping)))
```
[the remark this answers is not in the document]
vroldanbet: I get what you mean - we use `Must*` so that it lets us know when we've missed wiring up something in the code about a required flag. Typically it means we have a default value.

I'm ok turning this into an error, given it's better UX, but it's in conflict with the fact it needs to call `MustGetString` and thus needs to be named `Must*` too.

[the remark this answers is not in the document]
vroldanbet: addressed in https://github.com/authzed/zed/pull/316/commits/2028f7c23fcef35f2794fd14faa009e2a47c901d

josephschorr (approved):
LGTM

PR #321, opened 2023-12-29
Support validate (and import) taking in a schema directly

### LINKED ISSUE #287: Validate `.zaml` file should support sourcing schema from a file
Right now you need to copy/paste the schema from the `zed` file to `.zaml` file for validation. I would like there to be an option to just put the name of the file for the `schema:` key - that way I can avoid duplicating code.

josephschorr:
Fixes #287

### DIFF
--- a/internal/cmd/import.go
+++ b/internal/cmd/import.go
@@ -77,7 +77,7 @@ func importCmdFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	var p decode.SchemaRelationships
-	if _, err := decoder(&p); err != nil {
+	if _, _, err := decoder(&p); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


jzelinskie:
Most of the zed commands are "zed noun verb", but validation isn't.

WDYT about adding a "zed schema validate" and creating a new command for zed "zed file validate/import" for zedfiles.

Willing to bikeshed on the names eg "zed yaml validate".

josephschorr:
I don't really see a reason to break them into two commands: validate should work on both schema and extended files, especially because once https://github.com/authzed/spicedb/issues/1437 is implemented (or some variant thereof), there will likely be no distinction any longer

vroldanbet (approved):

PR #326, opened 2024-01-09
cmd: add autocomplete to `zed use`

jzelinskie:
This shortcut was missing the autocompletion available in the longform `zed context use`.


josephschorr (approved):
LGTM

PR #331, opened 2024-01-19
More and less completion

jzelinskie:
This PR does a few things
- consistently capitalizes usage text
- disables zsh's default file completion when not applicable
- uses file extensions for completion on schema/validation commands
- uses context completion for `zed schema copy`


josephschorr (approved):
LGTM

PR #340, opened 2024-02-01
Fix backup/restore per-second log output

alecmerdler:
Use `per_second` instead of `perSecond` to align with underscore_case used for the other log values.


alecmerdler:
We need to change all logging to use `camelCase`.

PR #343, opened 2024-02-09
fix ubuntu install instructions

vroldanbet:
if the file `/etc/apt/sources.list.d/fury.list` does not exist the command will fail with `Permission denied`


jzelinskie (approved):
Thanks! I tested this locally to confirm.

PR #355, opened 2024-03-18
add check bulk support in zed

vroldanbet:
the command look like
```
zed permission bulk <tuple1> <tuple2>
```

supports the json flag and consistency flags

The CLI accepts items as tuples so it can include a caveat context. The `--caveat-context` flag is not supported

### DIFF
--- a/internal/commands/permission.go
+++ b/internal/commands/permission.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"strings"
 
+	"github.com/authzed/spicedb/pkg/tuple"
+
 	"github.com/authzed/authzed-go/pkg/requestmeta"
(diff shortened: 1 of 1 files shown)


vroldanbet:
@jzelinskie do you have thoughts on the UX. The problem of accepting arguments like `resourcetype:resourceid relation objecttype:objectid` is that we need to represent the caveat as well, which until now was handled with`caveat--context`. The response is printed like: 

```
item => permissionship/error
item => permissionship/error
```

vroldanbet:
@jzelinskie friendly bump 🙏🏻

josephschorr (approved):
LGTM

PR #356, opened 2024-03-20
Allow passing request ID

alecmerdler:
Adds a `--request-id` flag to all `zed` commands. This improves the SpiceDB debugging experience by corrolating a specific request with SpiceDB server logs and OpenTelemetry traces.

### DIFF
--- a/internal/commands/util.go
+++ b/internal/commands/util.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/TylerBrock/colorjson"
+	"github.com/authzed/authzed-go/pkg/requestmeta"
 	"github.com/jzelinskie/cobrautil/v2"
 	"github.com/jzelinskie/stringz"
 	"github.com/spf13/cobra"
@@ -81,4 +82,16 @@ func PrettyProto(m proto.Message) ([]byte, error) {
 	}
 
 	return pretty, nil
+}
+
+// InjectRequestID adds the value of the --request-id flag to the
+// context of the given command.
+func InjectRequestID(cmd *cobra.Command, _ []string) error {
+	ctx := cmd.Context()
+	requestID := cobrautil.MustGetString(cmd, "request-id")
+	if ctx != nil && requestID != "" {
+		cmd.SetContext(requestmeta.WithRequestID(ctx, requestID))
+	}
+
+	return nil
 }
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/99designs/keyring v1.2.2
 	github.com/AlecAivazis/survey/v2 v2.3.7
 	github.com/TylerBrock/colorjson v0.0.0-20200706003622-8a50f05110d2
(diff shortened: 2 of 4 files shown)


alecmerdler:
@josephschorr Should we call this `--dev-request-header` to further clarify that this is only for debugging purposes?

internal/commands/permission.go
```diff
@@ -76,29 +76,34 @@ func RegisterPermissionCmd(rootCmd *cobra.Command) *cobra.Command {
 	checkCmd.Flags().Bool("schema", false, "requests debug information from SpiceDB and prints out the schema used")
 	checkCmd.Flags().Bool("error-on-no-permission", false, "if true, zed will return exit code 1 if subject does not have unconditional permission")
 	checkCmd.Flags().String("caveat-context", "", "the caveat context to send along with the check, in JSON form")
+	checkCmd.Flags().String("request-id", "", "optional id to send along with the request for tracing")
```
vroldanbet, line 79: Can't we turn this into a global flag so it does not have to be repeated for each subcommand?

internal/commands/permission.go
```diff
@@ -203,6 +213,10 @@ func checkCmdFunc(cmd *cobra.Command, args []string) error {
 		ctx = requestmeta.AddRequestHeaders(ctx, requestmeta.RequestDebugInformation)
 	}
 
+	if requestID != "" {
+		ctx = requestmeta.WithRequestID(ctx, requestID)
+	}
```
vroldanbet, line 218: to avoid the repetition, could we do this earlier in the stack? perhaps using cobras' lifecycle hooks?

vroldanbet (commented):
Left a few suggestions, I think we can DRY this a bit. I believe we already refer to this as `request-id` in the public docs so I'd suggest sticking to that name 👍🏻

internal/commands/permission.go
alecmerdler: Done

alecmerdler:
@vroldanbet Addressed your feedback, made `--request-id` a global flag which is injected into the context from a single lifecycle hook.

internal/commands/util.go
```diff
@@ -82,3 +83,14 @@ func PrettyProto(m proto.Message) ([]byte, error) {
 
 	return pretty, nil
 }
+
+// InjectRequestID adds the value of the --request-id flag to the
+// context of the given command.
+func InjectRequestID(cmd *cobra.Command, args []string) error {
+	requestID := cobrautil.MustGetString(cmd, "request-id")
```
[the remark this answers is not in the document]
alecmerdler: It doesn't panic because the default value is empty string.

internal/commands/util.go
```diff
@@ -82,3 +83,14 @@ func PrettyProto(m proto.Message) ([]byte, error) {
...
+	if requestID != "" {
+		cmd.SetContext(requestmeta.WithRequestID(cmd.Context(), requestID))
```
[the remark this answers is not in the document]
alecmerdler: Done

vroldanbet (approved):
Nice! LGTM! 🚀

PR #357, opened 2024-03-22
Respect resource ID filter when reading relationships

alecmerdler:
Fixes a bug in `zed relationship read` where the resource ID argument was not being applied, resulting in an incorrect filtering of relationships.

This regression was introduced by #341, which added filtering using prefixes.


josephschorr (approved):
LGTM

PR #354, opened 2024-03-18
support new `--explain` in 1.30

vroldanbet:

PR #358, opened 2024-03-22
Fix panic on permission check error

alecmerdler:
Ensures that we nil check the response from a CheckPermission request before attempting to extract the debug info from it.

This regression was introduced in #354 where we added support for the new debug info in the response body rather than in HTTP trailers.


vroldanbet (approved):

vroldanbet:
@alecmerdler feel free to release a patch version 🙏🏻

PR #365, opened 2024-04-02
updates bulk-delete to use DeleteRelationships new optional-limit

vroldanbet:
former implementation of bulk-delete was designed to support deleting large amounts of relationships in a batched way.

This commit replaces the implementation that relied on ReadRelationships with the new native DeleteRelationships `optional-limit`.

The contract is kept mostly intact, except the interactive prompt is removed as it was considered bad UX. Instead, the same limit of 1000 before the user is prompted is retained, and instead zed will fail, indicating the user to bump the limit with the new --optional-limit flag, or to use `--force`. **Non-zero error code is returned**.

The --optional-limit flag sets the corresponding `DeleteRelationship` request field, and `--force` now sets the `OptionalAllowPartialDeletions` field. If set to true, all relationships will be deleted in batches defined by `--optional-limit`, which continues to default to 1000 elements.

It also adds a progress bar to provide visual feedback to the user during deletion.


vroldanbet:
@josephschorr the wasm tests seem to be timing out in CI, but are passing locally. Do you have any ideas what could be going on?

internal/commands/relationship.go
```diff
@@ -47,10 +49,11 @@ func RegisterRelationshipCmd(rootCmd *cobra.Command) *cobra.Command {
 	registerConsistencyFlags(readCmd.Flags())
 
 	relationshipCmd.AddCommand(bulkDeleteCmd)
-	bulkDeleteCmd.Flags().Bool("force", false, "force deletion immediately without confirmation")
+	bulkDeleteCmd.Flags().Bool("force", false, "force deletion of all elements in batches defined by <optional-limit>")
 	bulkDeleteCmd.Flags().String("subject-filter", "", "optional subject filter")
+	bulkDeleteCmd.Flags().Uint("optional-limit", 1000, "the max amount of elements to delete. If you want to delete all in batches of size <optional-limit>, set --force to true")
 	bulkDeleteCmd.Flags().Bool("estimate-count", true, "estimate the count of relationships to be deleted")
-
+	_ = bulkDeleteCmd.Flags().MarkHidden("estimate-count")
```
josephschorr, line 56: Probably worth marking as deprecated and/or printing a warning if used

internal/commands/relationship.go
```diff
@@ -124,57 +127,62 @@ func bulkDeleteRelationships(cmd *cobra.Command, args []string) error {
...
+			RelationshipFilter:            filter,
+			OptionalLimit:                 uint32(optionalLimit),
+			OptionalAllowPartialDeletions: allowPartialDeletions,
+		}
+		log.Trace().Interface("request", delRequest).Msg("deleting relationships")
+
+		resp, err = spicedbClient.DeleteRelationships(cmd.Context(), delRequest)
+		if errorInfo, ok := grpcErrorInfoFrom(err); ok {
+			if errorInfo.GetReason() == v1.ErrorReason_ERROR_REASON_TOO_MANY_RELATIONSHIPS_FOR_TRANSACTIONAL_DELETE.String() {
+				return fmt.Errorf("partial deletion of type %s prevented, consider increasing --optional-limit beyond %s or deleting all relationships using --force",
```
josephschorr, line 149: I'd say something like: "could not delete %s, as found more than %d relationships"

internal/commands/relationship_test.go
```diff
@@ -549,3 +559,136 @@ func (m *mockClient) WriteRelationships(_ context.Context, in *v1.WriteRelations
 	require.True(m.t, proto.Equal(expectedWrite, in))
 	return &v1.WriteRelationshipsResponse{WrittenAt: &v1.ZedToken{Token: "test"}}, nil
 }
+
+func TestBulkDeleteForcing(t *testing.T) {
```
josephschorr, line 563: Can we add a test with, say, a couple hundred rels?

vroldanbet: @josephschorr done in https://github.com/authzed/zed/pull/365/commits/5f573aa5af5fd8474263516f04e41441c04836fd, PTAL 🙏🏻

internal/commands/relationship.go
```diff
@@ -124,57 +127,67 @@ func bulkDeleteRelationships(cmd *cobra.Command, args []string) error {
...
 
-		ctx, cancel := context.WithCancel(cmd.Context())
-		defer cancel()
+	allowPartialDeletions := cobrautil.MustGetBool(cmd, "force")
+	optionalLimit := cobrautil.MustGetUint(cmd, "optional-limit")
+	var resp *v1.DeleteRelationshipsResponse
+	for {
+		delRequest := &v1.DeleteRelationshipsRequest{
+			RelationshipFilter:            filter,
+			OptionalLimit:                 uint32(optionalLimit),
```
josephschorr, line 141: add a bounds check here

internal/console/console.go
```diff
@@ -24,3 +28,28 @@ func Println(values ...any) {
 		Printf("%v\n", value)
 	}
 }
+
+func CreateProgressBar(description string) *progressbar.ProgressBar {
```
josephschorr, line 32: doc comment

internal/commands/relationship.go
vroldanbet: addressed in https://github.com/authzed/zed/pull/365/commits/810734aa333c96bd9b7437c3965579293777885e

internal/console/console.go
vroldanbet: addressed in https://github.com/authzed/zed/pull/365/commits/a32b659634bd53a282ab9700bea8c70a2fac159c

josephschorr (approved):
LGTM

PR #367, opened 2024-04-04
introduces support for optional resource type in ResourceFilter

vroldanbet:
This introduces support for optional resource type, recently introduced in SpiceDB 1.30.0.

since the CLI contract has ambiguity, as a tradeoff, the contract will break by requiring clients to specify colon after the resource type, in order to disambiguate with _only specifying the relation_. The subject-type continues to not require colon as suffix.

Tests were also added for missing permutations and the optional resource id prefix.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,6 @@ toolchain go1.21.3
 
 require (
 	github.com/99designs/keyring v1.2.2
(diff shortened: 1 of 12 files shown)


andrii-harbour:
+1 to support feature request

PR #383, opened 2024-06-08
Add support for reading debug information from the error details

josephschorr:
Reference: https://github.com/authzed/spicedb/pull/1930

### DIFF
--- a/internal/commands/permission.go
+++ b/internal/commands/permission.go
@@ -20,6 +20,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/encoding/prototext"
 
 	"github.com/authzed/zed/internal/client"
 	"github.com/authzed/zed/internal/console"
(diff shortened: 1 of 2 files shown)


internal/commands/permission.go
```diff
@@ -222,8 +223,16 @@ func checkCmdFunc(cmd *cobra.Command, args []string) error {
...
+
+		// Check for the debug trace contained in the error details.
+		if errInfo, ok := grpcErrorInfoFrom(err); ok {
+			if encodedDebugInfo, ok := errInfo.Metadata["debug_trace_proto_text"]; ok {
+				debugInfo = &v1.DebugInformation{}
+				err = prototext.Unmarshal([]byte(encodedDebugInfo), debugInfo)
+				if err != nil {
+					return err
+				}
+			}
```
vroldanbet, line 235: Let's please add a test

vroldanbet (commented):
Changes look good to me, but to make sure it does not regress again, let's add a test.

internal/commands/permission.go
josephschorr: The main regression test is on the SpiceDB side but I'll see if I can add a test for parsing the proto here

josephschorr: Added

vroldanbet (approved):

PR #395, opened 2024-07-12
introduce lookup resources page limit

vroldanbet:
The new flag is `--page-limit`. If no limit is specified, it defaults to the original behavior, as it would be a breaking change (e.g. results that did not return duplicates now do).

Under the hood it uses cursoring


internal/commands/permission_test.go
```diff
@@ -98,3 +103,79 @@ func TestCheckErrorWithInvalidDebugInformation(t *testing.T) {
...
+		updates = append(updates, &v1.RelationshipUpdate{
+			Operation:    v1.RelationshipUpdate_OPERATION_TOUCH,
+			Relationship: tuple.ParseRel(fmt.Sprintf("test/resource:%d#reader@test/user:1", i)),
+		})
+	}
+
+	_, err = c.WriteRelationships(ctx, &v1.WriteRelationshipsRequest{Updates: updates})
+	require.NoError(t, err)
+
+	previous := console.Println
```
josephschorr, line 141: Add a comment here that you're overriding this

internal/commands/permission_test.go
```diff
@@ -98,3 +103,79 @@ func TestCheckErrorWithInvalidDebugInformation(t *testing.T) {
...
+	count = 0
+	cmd = testLookupResourcesCommand(t, 10)
+	err = lookupResourcesCmdFunc(cmd, []string{"test/resource", "read", "test/user:1"})
+	require.NoError(t, err)
+	require.Equal(t, 10, count)
+
+	// use odd page size
+	count = 0
+	cmd = testLookupResourcesCommand(t, 3)
+	err = lookupResourcesCmdFunc(cmd, []string{"test/resource", "read", "test/user:1"})
```
alecmerdler, line 166: It would be nice if we could validate that the parameter was sent to the server and that the responses are limiting the page size, but I'm unfamiliar with the test server code and whether or not that is possible right now.

alecmerdler (dismissed):
Nice addition!

internal/commands/permission_test.go
vroldanbet: https://github.com/authzed/zed/pull/395/commits/ebee0f0e092281d5bc39d428826028ddfb6b8224

internal/commands/permission_test.go
vroldanbet: That's a good point. I added some plumbing to be able to test this: https://github.com/authzed/zed/pull/395/commits/6a6c01a1881b6210295318bdeaa6ba97e4125b6b

alecmerdler: Nice!

alecmerdler (approved):

jzelinskie:
[Related to a Slack conversation](https://authzed.unthread.io/dashboard/inbox/all/conversations/6c787778-c390-4d71-a08e-7677136b46b0)

PR #400, opened 2024-07-17
zed backup redact: do not redact wildcards

vroldanbet:
wilcards were getting redacted, which then caused
a backup to fail to be restored, because the relationship written was not a wildcard.


pkg/backupformat/redaction.go
```diff
@@ -308,8 +308,11 @@ func redactRelationship(rel *v1.Relationship, redactionMap *RedactionMap, opts R
 
 	// Redact the object IDs.
 	if opts.RedactObjectIDs {
+		redactionMap.ObjectIDs["*"] = "*" // wilcards are not redacted
```
josephschorr, line 311: `tuple.PublicWildcard`

vroldanbet: done

josephschorr (approved):
LGTM

PR #401, opened 2024-07-17
backup redact: print objectID redaction mapping

vroldanbet:
so that customers can identify the mapping that
reproduces a specific issue


josephschorr (approved):
LGTM

PR #405, opened 2024-08-08
Add init that sets up logging for exceptional cases

### LINKED ISSUE #368: Some errors seem to always be in json, even if using an interactive shell
Example:
```sh
$ zed --insecure --token=test --endpoint=localhost:50051 lookup-subjects
{"level":"error","error":"unknown command \"lookup-subjects\" for \"zed\"","time":"2024-04-05T14:03:14-04:00","message":"terminated with errors"}
```

tstirrat15:
Fixes #368 

## Description
See #368. For error cases that were about missing commands or invalid argument sets, the logic that set up `zerolog` was never invoked, which meant that the output was a JSON blob instead of the prettily-formatted zerolog output. This makes the output more visually consistent.

## Changes
* Add `init` to command that sets up logging in the event that `zl.RunE` isn't actually called.
## Testing
Review. Checkout, `go build cmd/zed/..`, and run the resulting binary with some invalid args. `./main foo` and `./main validate` should do it.

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -3,10 +3,12 @@ package cmd
 import (
 	"context"
 	"errors"
+	"io"
 	"os"
 	"os/signal"
 	"syscall"
 
+	"github.com/mattn/go-isatty"
 	"github.com/jzelinskie/cobrautil/v2"
 	"github.com/jzelinskie/cobrautil/v2/cobrazerolog"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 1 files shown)


tstirrat15:

internal/cmd/cmd.go
```diff
@@ -21,6 +23,25 @@ var (
...
 
+func init() {
+	// NOTE: this is mostly to set up logging in the case where
+	// the command doesn't exist or the construction of the command
+	// errors out before the PersistentPreRunE setup in the below function.
+	// It helps keep log output visually consistent for a user even in
+	// exceptional cases.
+	var output io.Writer
+
+	if isatty.IsTerminal(os.Stdout.Fd()) {
```
ecordell, line 34: should this check `Stderr` since that's where we're writing to?

tstirrat15: I cribbed from the [cobrazerolog](https://github.com/jzelinskie/cobrautil/blob/main/cobrazerolog/cobrazerolog.go#L93-L98) source, so I'll update it there if necessary.

I think that would technically be safer, but I also can't think of a situation where you'd have a TTY file descriptor on stdout but not stderr.

ecordell: @jzelinskie thoughts here since you wrote this in cobrazerolog?

I wouldn't block on this but if it makes sense to change we should change both places.

ecordell (approved):

PR #407, opened 2024-08-28
Add option for max message size

tstirrat15:
## Description
We had a user run into an issue where `zed backup <filename>` failed with:
```
2:08PM ERR terminated with errors error="error receiving relationships: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (6280392 vs. 4194304)"
```
This indicates that the client received a 6mb message with a max configured message size of 4mb, which is the default for a gRPC client.

This seems like something that could reasonably happen with a sufficiently large schema, especially if it has lots of comments.

My understanding is that this is a securityish feature meant to prevent a client from choking on an incoming message. It would also prevent a client from getting OOMkilled as a result of attempting to read a too-large message into memory. For the zed client, it seems like this isn't a concern, since it would typically be running on either a user's device or a CI server, neither of which are particularly memory-constrained.

## Changes
* Add a command line flag for dial opts that increase the max message size for both send and receive
## Testing
Review. See that tests still pass.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -91,19 +91,9 @@ func DialOptsFromFlags(cmd *cobra.Command, token storage.Token) ([]grpc.DialOpti
 		interceptors = append(interceptors, zgrpcutil.CheckServerVersion)
 	}
 
-	maxMessageSize := cobrautil.MustGetInt(cmd, "max-message-size")
-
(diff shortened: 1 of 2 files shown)


josephschorr (approved):
LGTM

PR #420, opened 2024-09-19
Fix CLI usage with no pre-set context/config

tstirrat15:
## Description
In #417 we introduced a bug where `zed` became unusable if there wasn't previously a context set by the user. This wasn't intentional; it should be valid to invoke zed entirely from the command line in isolation. This PR captures that behavior in a unit test and fixes the behavior.

## Changes
* Separate existence and fetching concerns for tokens
* Handle cases where config file or token don't exist
## Testing
See that tests go green. See that you can clear out your local context (or run from within a container): `/zed schema write ./blah.zed --endpoint 127.0.0.1:50051 --token supersecretdevkey --insecure` and not receive an error.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -74,6 +74,14 @@ func newClientForContext(cmd *cobra.Command, contextName string, secretStore sto
 
 // GetCurrentTokenWithCLIOverride returns the current token, but overridden by any parameter specified via CLI args
 func GetCurrentTokenWithCLIOverride(cmd *cobra.Command, configStore storage.ConfigStore, secretStore storage.SecretStore) (storage.Token, error) {
+	// Handle the no-config case separately
+	configExists, err := configStore.Exists()
+	if err != nil {
+		return storage.Token{}, err
+	}
+	if !configExists {
+		return GetTokenWithCLIOverride(cmd, storage.Token{})
+	}
 	token, err := storage.CurrentToken(
 		configStore,
 		secretStore,
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -54,7 +54,7 @@ func newClientForCurrentContext(cmd *cobra.Command) (Client, error) {
 }
 
 func newClientForContext(cmd *cobra.Command, contextName string, secretStore storage.SecretStore) (*authzed.Client, error) {
-	currentToken, err := storage.GetToken(contextName, secretStore)
+	currentToken, err := storage.GetTokenIfExists(contextName, secretStore)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


internal/storage/config.go
```diff
@@ -139,6 +148,17 @@ func (s JSONConfigStore) Put(cfg Config) error {
...
 
+func (s JSONConfigStore) Exists() (bool, error) {
+	_, err := os.Stat(filepath.Join(s.ConfigPath, configFileName))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
```
ecordell, line 159: purely a style nit, but I'm used to seeing this written more like:
```suggestion
	if _, err := os.Stat(filepath.Join(s.ConfigPath, configFileName)); errors.Is(err, fs.ErrNotExist) {
	} else if err != nil {
	} else {
	}
```

also apparently `IsNotExist` is soft-deprecated in favor of `errors.Is`: https://github.com/golang/go/blob/master/src/os/error.go#L88-L89

internal/storage/secrets.go
```diff
@@ -84,7 +82,22 @@ func GetToken(name string, ss SecretStore) (Token, error) {
...
+	return Token{}, nil
+}
+
+func TokenExists(name string, ss SecretStore) (bool, error) {
+	secrets, err := ss.Get()
+	if err != nil {
+		return false, err
+	}
+
+	for _, token := range secrets.Tokens {
```
ecordell, line 94: this is fine but makes me think `Tokens` should just be a `map[string]Token` instead of `[]Token`

Also maybe this would make sense as a method on `SecretStore`?

ecordell (approved):
left some comments, but non-blocking

internal/client/client.go
```diff
@@ -74,6 +74,14 @@ func newClientForContext(cmd *cobra.Command, contextName string, secretStore sto
 
 // GetCurrentTokenWithCLIOverride returns the current token, but overridden by any parameter specified via CLI args
 func GetCurrentTokenWithCLIOverride(cmd *cobra.Command, configStore storage.ConfigStore, secretStore storage.SecretStore) (storage.Token, error) {
+	// Handle the no-config case separately
+	configExists, err := configStore.Exists()
+	if err != nil {
+		return storage.Token{}, err
+	}
+	if !configExists {
+		return GetTokenWithCLIOverride(cmd, storage.Token{})
```
josephschorr, line 83: shouldn't CLI override... override the context?

internal/storage/config.go
tstirrat15: I dig

internal/storage/secrets.go
tstirrat15: Yeah, as I was going through this there were a bunch of things that seemed like they could be reorganized. I want to get the fix out and then loop back.

tstirrat15:
Follow-on work around code organization can happen in the context of #421.

josephschorr (approved):
LGTM

PR #431, opened 2024-10-30
Update against most recent spicedb

tstirrat15:
## Description
This is mostly prepping for SpiceDB 1.38 and to unblock other work on zed that I wanted to do with a local copy of the most recent version of SpiceDB. The changes that we pushed through to the core structs needed to be reflected in their usages on the zed side.

## Note
I think this should be safe to merge even before the 1.38 changes go out, since the SpiceDB internals imported here are mostly used for schema validation and manipulation.

## Changes
* Pick up most recent version of authzed-go for API changes in SpiceDB
* Pick up most recent version of SpiceDB for internal struct changes
* Fix usages of deprecated struct functions
## Testing
Review. See that things go green.


internal/cmd/backup_test.go
```diff
@@ -300,12 +300,13 @@ func TestBackupCreateCmdFunc(t *testing.T) {
 	_, err = c.WriteSchema(ctx, &v1.WriteSchemaRequest{Schema: testSchema})
 	require.NoError(t, err)
 
-	testRel := "test/resource:1#reader@test/user:1"
+	testRelString := "test/resource:1#reader@test/user:1"
+	testRel, _ := tuple.ParseV1Rel(testRelString)
```
tstirrat15, line 304: I'm gonna go back through these and use `tuple.MustParse().ToCoreTuple()` instead.

tstirrat15: Nvm it's a different type.

.github/workflows/build-test.yaml
```diff
@@ -1,10 +1,6 @@
 ---
 name: "Build & Test"
 on: # yamllint disable-line rule:truthy
-  push:
-    branches:
-      - "!dependabot/*"
-      - "*"
```
tstirrat15, line 7: This was double-running all of our tests. If we still need to do something different with dependabot we can address that separately.

pkg/wasm/main_test.go
```diff
@@ -69,3 +69,9 @@ func TestZedCommand(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, updatedContext.Relationships, 3)
 }
+
+// NOTE: this is required to preserve caveat context.
+// tuple.CoreRelationToString elides the caveat context.
+func stringifyTuple(relationtuple *v1.RelationTuple) string {
+	return tuple.MustString(tuple.FromCoreRelationTuple(relationtuple))
```
josephschorr, line 76: Remove

josephschorr (approved):
LGTM

PR #432, opened 2024-10-30
Add zed schema compile command

tstirrat15:
## Description
This is the first pass of the `zed schema compile` command. It adds it under a `preview` subcommand so that we can release to the public without setting an expectation that the API is stable. This should allow us to do releases and get feedback without locking ourselves onto a particular course.

## Changes
* Add workfiles to gitignore
* Add readme section on difference between `cmd` and `commands`
* Add preview command for schema compilation

## Testing
Run `go build ./cmd/zed/...` and then run `./zed preview schema compile --help` and see that it's working


go.work
tstirrat15, line 1: This'll be removed before merge.

go.work.sum
tstirrat15, line 1: This'll be removed before merge.

tstirrat15:
Evan pointed out that absolute paths don't seem to work, and that seems like a sane use case. I'll fix that.

.gitignore
```diff
@@ -1,5 +1,12 @@
 dist/
+temp/
 
 # Local-only files
 go.work
 go.work.sum
+
```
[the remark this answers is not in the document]
tstirrat15: Yeah I still haven't figured out merge conflict resolution with jujutsu :sweat_smile:

ecordell (approved):
LGTM

Not blocking, but I do think the error message could be better when the file being referenced is missing:

```
9:18AM ERR terminated with errors error="failed to read schema file: open /Users/evan/Downloads/schemastesting/subjects.zed: no such file or directory"
```

Ideally this would say something about the `import` statement on line 1 of `root.zed`, etc.

PR #435, opened 2024-11-07
Add --force-color flag to zed validate

### LINKED ISSUE #434: Option to Force Color Output in zed validate Command
It would be nice to have an option in the zed validate command to force color output, regardless of the TTY status. The colorized output is really helpful in highlighting warnings and other important details. 

Some CI providers, such as BuildKite can display annotations with ANSI color codes. Currently the color output is lost because the output is not going directly to a TTY. It appears the underlying library, [LipGloss](https://github.com/charmbracelet/lipgloss?tab=readme-ov-file#faq), automatically disables color output when it detects a non-TTY environment, but it also allows overriding this behavior. 

**Proposed Solution:**

Add a --force-color flag (or similar) to zed validate to allow color output to be explicitly enabled.

tstirrat15:
Fixes #434 

## Description
See #434. There are non-terminal environments (e.g. some CI environments) that support color, but `lipgloss` assumes that a non-terminal environment is uncolored. This provides a `--force-color` flag that allows the user to override this behavior and get color output for `zed validate`, which is the only place we're using `lipgloss`.

## Changes
* Add flag for color forcing
* Add `PreRunE` that interprets the flag
* Thunk the functions that produce output so that they correctly pick up the changes to the `DefaultRenderer`
* Propagate the thunks

## Testing
Review. See that things are green. The way that I was able to test this locally was comparing:

```sh
docker run -v './zed:/zed' -v '/etc/ssl/certs:/etc/ssl/certs' busybox ./zed validate https://gist.github.com/ecordell/8e3b613a677e3c844742cf24421c08b6

docker run -v './zed:/zed' -v '/etc/ssl/certs:/etc/ssl/certs' busybox ./zed validate https://gist.github.com/ecordell/8e3b613a677e3c844742cf24421c08b6 --force-color
```

Where the command is run from a folder that has a built `zed` binary.

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -84,7 +84,6 @@ func validatePreRunE(cmd *cobra.Command, _ []string) error {
 	setForceColor := cobrautil.MustGetBool(cmd, "force-color")
 	if setForceColor {
 		lipgloss.SetColorProfile(termenv.ANSI256)
-		fmt.Println(lipgloss.DefaultRenderer().ColorProfile() == termenv.ANSI256)
 	}
 
 	return nil
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -16,6 +16,8 @@ import (
 	"github.com/authzed/spicedb/pkg/spiceerrors"
 	"github.com/authzed/spicedb/pkg/validationfile"
 	"github.com/charmbracelet/lipgloss"
+	"github.com/jzelinskie/cobrautil/v2"
+	"github.com/muesli/termenv"
 
 	"github.com/authzed/zed/internal/commands"
 	"github.com/authzed/zed/internal/console"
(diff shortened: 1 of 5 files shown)


internal/cmd/validate.go
```diff
@@ -64,9 +73,23 @@ var validateCmd = &cobra.Command{
...
 }
 
+func validatePreRunE(cmd *cobra.Command, _ []string) error {
+	// Override lipgloss's autodetection of whether it's in a terminal environment
+	// and display things in color anyway. This can be nice in CI environments that
+	// support it.
+	setForceColor := cobrautil.MustGetBool(cmd, "force-color")
+	if setForceColor {
+		lipgloss.SetColorProfile(termenv.ANSI256)
+		fmt.Println(lipgloss.DefaultRenderer().ColorProfile() == termenv.ANSI256)
```
vroldanbet, line 87: why the need to print here?

internal/cmd/validate.go
```diff
@@ -24,21 +26,28 @@ import (
...
-	errorMessageStyle      = lipgloss.NewStyle().Bold(true).Width(80)
-	linePrefixStyle        = lipgloss.NewStyle().Foreground(lipgloss.Color("12"))
-	highlightedSourceStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("9"))
-	highlightedLineStyle   = lipgloss.NewStyle().Foreground(lipgloss.Color("9"))
-	codeStyle              = lipgloss.NewStyle().Foreground(lipgloss.Color("8"))
-	highlightedCodeStyle   = lipgloss.NewStyle().Foreground(lipgloss.Color("15"))
-	traceStyle             = lipgloss.NewStyle().Bold(true)
+	// NOTE: these need to be set *after* the renderer has been set, otherwise
+	// the forceColor setting can't work, hence the thunking.
+	success = func() string {
```
vroldanbet, line 31: what's the cost of initializing this each time? alternatively, these could be left as uninitialized vars, and initialize them after we've checked the flag

internal/cmd/validate.go
tstirrat15: Good catch :facepalm:

internal/cmd/validate.go
tstirrat15: I thought about that, but that spreads the definitions around the file, and some of the definitions are used by multiple functions. I don't think it's particularly expensive, though that isn't because I've profiled it.

vroldanbet (approved):

ecordell (approved):

PR #444, opened 2024-12-26
Display usage string when incorrect arg number

### LINKED ISSUE #443: Incorrect arg number should print the help for that command
I often forget which arguments are required for certain commands - if you mess this up, the error message doesn't give you any help at all.

Example:
```
$ zed relationship  read
9:07PM ERR terminated with errors error="accepts between 1 and 3 arg(s), received 0"
```

Instead, it would be nice if we printed the help text for that command as well, like:
```
$ zed relationship  read -h
9:07PM ERR terminated with errors error="accepts between 1 and 3 arg(s), received 0"

Usage:
  zed relationship read <resource_type:optional_resource_id> <optional_relation> <optional_subject_type:optional_subject_id#optional_subject_relation> [flags]
  ...
```

(I believe this is just a flag in cobra that we're not setting)

kartikaysaxena:
Fixes #443 

Screenshot 2024-12-26 at 8 39 29 PM


kartikaysaxena:

kartikaysaxena:
recheck

tstirrat15 (approved):
Yep, this looks good to me. Thank you!

PR #450, opened 2025-01-07
Document resource ID prefix filtering

alecmerdler:
Adds an example to the help output for `zed relationship read` with an example of how to filter the results using a resource ID prefix.

**New output:**
```
$ zed relationship read --help
Enumerates relationships matching the provided pattern.

To filter returned relationships using a resource ID prefix, append a '%' to the resource ID:

zed relationship read some-type:some-prefix-%

Usage:
  zed relationship read <resource_type:optional_resource_id> <optional_relation> <optional_subject_type:optional_subject_id#optional_subject_relation> [flags]
```


tstirrat15 (approved):
LGTM!

PR #452, opened 2025-01-08
Add --explain support for check bulk permission

josephschorr:
1) Now supports `--explain` on check bulk permission (using HEAD of SpiceDB)
2) Adds support for combined state display on batched permission checks
3) Adds source support for distinguishing which cache is used (when applicable)


tstirrat15 (dismissed):
LGTM

tstirrat15 (approved):

PR #456, opened 2025-01-17
Added magefile to generate docs with github workflow

### LINKED ISSUE #234: Make docs for cli

kartikaysaxena:
Fixes #234 

Added github workflow that would invoke mage to generate docs for zed and submit the changes to the docs repo.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,85 @@
+---
+name: "Sync Generated Docs"
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - "main"
+
+env:
+  DOCS_REPO: "authzed/docs"
+  DOCS_BRANCH: "main"
+  GENERATED_DOCS_DIR: "docs"
+  TARGET_DOCS_DIR: "pages/zed"
+
+permissions:
+  contents: "write"
+  pull-requests: "write"
+
+jobs:
+  generate-and-sync-docs:
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: "Checkout source repository"
+        uses: "actions/checkout@v3"
+        with:
+          fetch-depth: 1
+
+      - name: "Set up Go"
+        uses: "actions/setup-go@v4"
+        with:
+          go-version: 1.20
+
+      - name: "Generate documentation"
+        run: |
+          cd magefiles
+          if ! mage gen:docs; then
+            echo "Documentation generation failed"
+            exit 1
+          fi
+
+      - name: "Clone docs repository"
+        run: |
+          git clone --depth 1 --branch $DOCS_BRANCH https://github.com/$DOCS_REPO.git docs-repo || {
+            echo "Failed to clone docs repository"
+            exit 1
+          }
+
+      - name: "Compare generated docs with target docs"
+        id: "compare"
+        run: |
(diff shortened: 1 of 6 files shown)


magefiles/magefile.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"os"
+
+	"github.com/authzed/zed/internal/cmd"
+	"github.com/jzelinskie/cobrautil/v2/cobrazerolog"
+	"github.com/spf13/cobra/doc"
+)
+
+// Generate markdown files for zed.
+func GenerateDocs() error {
```
josephschorr, line 15: Mind naming this `gen:docs` and adding a `gen:all` as well?

We use that form in SpiceDB for any generated code: https://github.com/authzed/spicedb/blob/main/magefiles/gen.go#L16

josephschorr: This would also be great to add to SpiceDB as a followup

internal/cmd/cmd.go
```diff
@@ -42,9 +42,7 @@ func init() {
 	log.Logger = l
 }
 
-func Run() {
-	zl := cobrazerolog.New(cobrazerolog.WithPreRunLevel(zerolog.DebugLevel))
-
+func InitialiseRootCmd(zl *cobrazerolog.Builder) *cobra.Command {
```
josephschorr, line 45: Add a doc comment on this exported method

magefiles/magefile.go
kartikaysaxena: sure

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+      - name: "Set up Go"
+        uses: "actions/setup-go@v4"
+        with:
+          go-version: 1.20
+
+      - name: "Generate documentation"
+        run: |
+          cd magefiles
+          if ! mage GenerateDocs; then
```
josephschorr, line 35: Fix command

kartikaysaxena: sure

josephschorr:
@kartikaysaxena Mind opening a PR with the docs on the docs repo so I can see them as well?

kartikaysaxena:
> @kartikaysaxena Mind opening a PR with the docs on the docs repo so I can see them as well?

Here it is authzed/docs#296

kartikaysaxena:
Also authzed/spicedb#2208 and authzed/docs#298 for SpiceDB docs

kartikaysaxena:
Another way of doing this is to use commands like `spicedb generate-docs` and `zed generate-docs` to generate ingestable docs for other tools to use, not exposing them in docs. What do you think @josephschorr

jzelinskie:
I think this is wayyy too many files for our docs.
It'd be great if we could consolidate it into one regular docs page.

Another route to accomplish this could be reusing our existing manpage and pandoc to generate markdown.

kartikaysaxena:
@jzelinskie 
Merged into a single doc and looks like [this](https://gist.github.com/kartikaysaxena/63c0707e1ea1f852b6f96ffd627d97de). Does it need more improvement?

jzelinskie:
Looks great to me!

tstirrat15 (approved):
Looks good to me! Let's get this in.

tstirrat15:
Thank you for putting this up!

PR #458, opened 2025-01-27
Added expiration functionality to relationship writing

### LINKED ISSUE #448: Add expiration functionality to relationship writing
I haven't actually tested this yet, but I'm assuming that we'll need to modify the CLI to allow for writing relationships that have expirations.

kartikaysaxena:
Fixes #448 added a flag `--expiration-time` to take the input timestamp in RFC 3339 format, also added tests.


internal/commands/relationship_test.go
tstirrat15, line 1: Please separate out tests that make assertions about expiration behavior and leave the existing tests as they are.

internal/commands/relationship.go
```diff
@@ -491,7 +495,7 @@ func writeRelationshipCmdFunc(operation v1.RelationshipUpdate_Operation, input *
 			}
 
 			if operation != v1.RelationshipUpdate_OPERATION_DELETE {
-				if err := handleCaveatFlag(cmd, rel); err != nil {
+				if err := handleCaveatAndExpirationFlags(cmd, rel); err != nil {
 					return err
 				}
```
tstirrat15, line 500: I'd prefer to handle the different flags inline here - remove `handleCaveatAndExpirationFlags` and have a separate `if` statement here for `handleExpiration`.

tstirrat15 (changes requested):
See comments - let me know if you have any questions on them.

kartikaysaxena:
It seems like there are some vulnerabilities with version of Go in goreleaser which trivy is complaining while analysing the built images

tstirrat15 (approved):
Nice, thank you

Issue #464, opened 2025-02-06
workflow: fix docs workflow


kartikaysaxena:
Didn't fix the workflow while #456 was merged, right now would be failing for every PR

tstirrat15:
Yeah, my bad on that - I didn't check the workflow before I merged it. I think it will help to use both the `setup-go` and `setup-mage` actions from this repo: https://github.com/authzed/actions

That should keep the go and mage versions consistent with the other repositories in the organization.

kartikaysaxena:
My bad, just tested new changes on my fork https://github.com/kartikaysaxena/zed/actions/runs/13202513070/job/36857732133 which resulted this PR https://github.com/kartikaysaxena/docs/pull/1

miparnisari:
The workflow to publish to the `docs` repo works now. What's left is to decide what changes we want to send over - i.e. https://github.com/authzed/zed/pull/503

PR #466, opened 2025-02-07
docs: fixed workflow

kartikaysaxena:
Related #464 
Fixed workflow, tested on fork, example PR https://github.com/kartikaysaxena/docs/pull/1
The token should have read/write permissions

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -15,6 +15,8 @@ env:
 permissions:
   contents: "write"
   pull-requests: "write"
+  actions: "write"
+  repository-projects: "write"
 
 jobs:
   sync-docs:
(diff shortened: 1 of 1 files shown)


tstirrat15 (commented):
See comment - otherwise it's looking good to me.

.github/workflows/docs.yaml
```diff
@@ -30,48 +32,42 @@ jobs:
...
+            gh auth login --with-token < token.txt
+            gh pr create \
+              --body "" \
+              --title "docs: update zed docs" \
+              --head "update-generated-docs" \
+              --base "main"
+          else
+            echo "No changes detected in $TARGET_DOCS_FILE."
+            echo "CHANGES_DETECTED=false" >> $GITHUB_ENV
+          fi
```
tstirrat15, line 73: Have you seen this action? https://github.com/marketplace/actions/create-pull-request

Would it work for this use case? We use it in other actions: https://github.com/authzed/authzed-node/blob/main/.github/workflows/automatic-api-update.yaml#L37-L47

tstirrat15 (commented):
Sorry, this is the comment I meant to leave

.github/workflows/docs.yaml
kartikaysaxena: Yeah did have a look over the documentation, https://github.com/kartikaysaxena/docs/pull/2 was the result of changes, switching to this action.

tstirrat15:
There was a perms failure: https://github.com/authzed/zed/actions/runs/13243163183/job/36963066495

I think there might be something about a PAT or the permissions on the token declared at the top of the file? I'm curious as to why it worked in your fork but not here.

PR #467, opened 2025-02-10
workflow: perms updated

kartikaysaxena:
I used these perms while debugging the same 403 with an earlier approach, `peter-evans/create-pull-request` action didn't need these perms on my fork (something to do with gh token), should work this time, checked with this just now https://github.com/kartikaysaxena/zed/actions/runs/13248365111/workflow


tstirrat15 (approved):
Cool, let's give it a go

tstirrat15:
Same issue :/

PR #470, opened 2025-02-24
feat: json output for schema in `zed preview schema compile`

### LINKED ISSUE #469: Allow JSON output from file input
It would be helpful to be able to utilize the CLI to get a json representation of a schema even without a SpiceDB instance available.

We are already have the ability to validate a schema from a file, but having the json representation available for consumption by other tooling would greatly enhance our current workflows.

kartikaysaxena:
Fixes #469

This introduces a `json` flag to view the compiled schema in json format without the need of a live SpiceDB deployment, also can be written with an additional `out` flag.


miparnisari:
@kartikaysaxena do you mind adding one unit test?

PR #478, opened 2025-03-11
Add support for composable schemas in `zed validate`

tstirrat15:
## Description
Part of composable schema support. We want to make it so that `zed validate` can operate on both composable schemas and standard schemas.

## Changes

* By default, `validate` will attempt to validate a schema first according to composable schema rules, then standard schema rules, and if both fail it will show the errors from composable schema.
* There's a new flag: `schema-syntax`. If you add `--schema-syntax composable`, it will validate according to only composable schema syntax rules, and if you add `--schema-syntax standard`, it will validate according to only standard schema rules. The default value for this flag is the empty string.

## Testing
- Manual testing to ensure no regressions were introduced
- Unit tests

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -13,9 +13,11 @@ import (
 	"github.com/muesli/termenv"
 	"github.com/spf13/cobra"
 
+	composable "github.com/authzed/spicedb/pkg/composableschemadsl/compiler"
 	"github.com/authzed/spicedb/pkg/development"
 	core "github.com/authzed/spicedb/pkg/proto/core/v1"
 	devinterface "github.com/authzed/spicedb/pkg/proto/developer/v1"
+	"github.com/authzed/spicedb/pkg/schemadsl/compiler"
 	"github.com/authzed/spicedb/pkg/spiceerrors"
 	"github.com/authzed/spicedb/pkg/validationfile"
 
@@ -47,6 +49,7 @@ var (
 
 func registerValidateCmd(cmd *cobra.Command) {
 	validateCmd.Flags().Bool("force-color", false, "force color code output even in non-tty environments")
+	validateCmd.Flags().String("schema-type", "", "force validation according to specific schema syntax (\"\", \"composable\", \"standard\")")
 	cmd.AddCommand(validateCmd)
 }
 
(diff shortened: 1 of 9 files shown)


internal/cmd/validate.go
```diff
@@ -340,3 +387,30 @@ func renderLine(sb *strings.Builder, lines []string, index int, highlight string
 			highlightedSourceStyle().Render(strings.Repeat("~", highlightLength)))
 	}
 }
+
+// classifyErrors returns errors from the composable DSL, the standard DSL, and any other parsing errors.
+func classifyErrors(err error) (error, error, error) {
+	if err == nil {
+		return nil, nil, nil
+	}
+	var (
```
josephschorr, line 396: I prefer a var-per-line, but that's me

internal/decode/decoder.go
```diff
@@ -166,22 +170,65 @@ func unmarshalAsYAMLOrSchema(data []byte, out interface{}) (bool, error) {
...
+		}
+		vfile.Schema = blocks.ParsedSchema{
+			CompiledSchema: &compiler.CompiledSchema{
+				ObjectDefinitions:  composableCompiled.ObjectDefinitions,
+				CaveatDefinitions:  composableCompiled.CaveatDefinitions,
+				OrderedDefinitions: oldDefinitions,
+			},
+			Schema:         compiledSchemaString,
+			SourcePosition: spiceerrors.SourcePosition{LineNumber: 1, ColumnPosition: 1},
+		}
```
tstirrat15, line 227: This feels like an interim solution to me - the shapes happen to match, but that may not continue to be the case as each compiler package evolves, and we'll have deferred this work until then.

Was that an intentional decision? I'm fine with an answer of "yes", especially given that when I tried to change the validationfile package to use an interface, I disappeared down a rabbit hole for a week.

If it was, can we add a comment to that effect?

tstirrat15 (commented):
Copying the definitions around feels like a clever way to get around the types differing, but I'm a little worried about it.

internal/decode/decoder.go
miparnisari: So, the reason i'm doing this is that the `validate` code uses `var parsed validationfile.ValidationFile` and I didn't want to change that. 

And so, I borrowed code from https://github.com/authzed/zed/blob/d3826bf5872bd3722fdf46b85b6ecc05c89d6365/internal/cmd/preview.go#L90-L101

I guess, looking more in detail, I only need the `compiledSchemaString` 🤔

tstirrat15 (commented):
Generating the schema string for composable schemas seems like an elegant solution. Looks good to me!

miparnisari (approved):

Issue #482, opened 2025-03-31
Support cursor parameter for zed permission lookup-resources


kbrwn:
### Description
The `zed`cli currently doesn't provide a way to specify a cursor parameter when using `permission lookup-resources`. While the CLI supports pagination through the `--page-limit` flag, there's no way to manually resume pagination from a specific cursor position or to share a pagination state between different command invocations.

### Desired Behavior
- Add a `--cursor` flag to relevant commands (e.g., `lookup-resources`)
- Support providing a cursor token as a parameter to resume pagination
- Support outputting the last cursor with an option like `--show-cursor`

### Example CLI Usage (Proposed)
```
# Get first page and show the cursor
zed permission lookup-resources document view user:user1 --page-limit 100 --show-cursor

# Resume from a specific cursor
zed permission lookup-resources document view user:user1 --page-limit 100 --cursor "MzQxODI5Mzc0ODM2NDgwMA=="

miparnisari:
This was done in https://github.com/authzed/zed/pull/485.

PR #485, opened 2025-04-01
perms: add support for cursor and show-cursor flags in lookup subjects

kartikaysaxena:
Related #482

Adds support for displaying the last cursor when the user opts in a `show-cursor`, also allows user to use a custom cursor to resume browsing the pages from a specific state

### DIFF
--- a/internal/commands/permission.go
+++ b/internal/commands/permission.go
@@ -107,6 +107,8 @@ func RegisterPermissionCmd(rootCmd *cobra.Command) *cobra.Command {
 	lookupResourcesCmd.Flags().String("revision", "", "optional revision at which to check")
 	lookupResourcesCmd.Flags().String("caveat-context", "", "the caveat context to send along with the lookup, in JSON form")
 	lookupResourcesCmd.Flags().Uint32("page-limit", 0, "limit of relations returned per page")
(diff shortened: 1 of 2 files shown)


internal/commands/permission.go
```diff
@@ -521,6 +527,11 @@ func lookupResourcesCmdFunc(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	showCursor := cobrautil.MustGetBool(cmd, "show-cursor")
+	if showCursor && cursor != nil {
+		console.Printf("Last cursor: %s\n", cursor.Token)
+	}
```
tstirrat15, line 533: Can we change the logic to set `show-cursor` to true by default, and show the cursor if one exists by default?

I think it'll be more useful to show it if it's available, unless the user explicitly says not to. I chatted with the issue reporter about this and we're in agreement there.

tstirrat15 (commented):
See comments - I'll re-review after that

internal/commands/permission.go
kartikaysaxena: sure, would be handy, defaulting the flag to be true unless explicitly set to false

tstirrat15 (approved):
Looks good to me! Thank you!

tstirrat15:
That test looks like a potential flake - I'll look into it if it fails again

PR #491, opened 2025-04-21
add more validate tests

miparnisari:
In preparation for https://github.com/authzed/zed/pull/478, adding more tests to ensure we don't break existing behavior.

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -74,7 +74,17 @@ var validateCmd = &cobra.Command{
 	Args:              cobra.MinimumNArgs(1),
 	ValidArgsFunction: commands.FileExtensionCompletions("zed", "yaml", "zaml"),
 	PreRunE:           validatePreRunE,
-	RunE:              validateCmdFunc,
+	RunE: func(cmd *cobra.Command, filenames []string) error {
+		result, shouldExit, err := validateCmdFunc(cmd, filenames)
+		if err != nil {
+			return err
+		}
+		console.Print(result)
+		if shouldExit {
+			os.Exit(1)
+		}
+		return nil
+	},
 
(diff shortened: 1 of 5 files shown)


internal/cmd/validate.go
```diff
@@ -94,37 +104,41 @@ func validatePreRunE(cmd *cobra.Command, _ []string) error {
...
+			return "", false, err
 		}
 
 		var parsed validationfile.ValidationFile
 		validateContents, isOnlySchema, err := decoder(&parsed)
 		if err != nil {
 			var errWithSource spiceerrors.WithSourceError
 			if errors.As(err, &errWithSource) {
-				ouputErrorWithSource(validateContents, errWithSource)
+				outputErrorWithSource(toPrint, validateContents, errWithSource)
```
miparnisari, line 138: Not sure how to hit this code path yet

internal/cmd/validate.go
```diff
@@ -151,115 +165,116 @@ func validateCmdFunc(cmd *cobra.Command, filenames []string) error {
 			}
 
 			// Output errors
-			outputDeveloperErrorsWithLineOffset(validateContents, devErrs.InputErrors, schemaOffset)
+			outputDeveloperErrorsWithLineOffset(toPrint, validateContents, devErrs.InputErrors, schemaOffset)
```
miparnisari, line 168: Not sure how to hit this code path yet

tstirrat15: IIRC this and the other are some combination of compilation errors and typesystem errors. Grepping for `DeveloperErr` and `ErrorWithSource` should probably turn something up.

tstirrat15: JK you've already got some of that kind of thing in there...

tstirrat15 (approved):
I like where this is headed

PR #492, opened 2025-04-22
introduce resumable backups

vroldanbet:
This PR modifies the backup create command to allow an incomplete backup to be resumed.

A new marker file is added with the last written bulk export cursor. This was not added to the OCF file because OCF container is not meant for in-place updates, but streaming append operations. The marker is updated every time a bulk export page is successfully written.

For convenience, the command now also supports creating a backup without a file name; in this case, it will derive the backup file name from the Zed context name. A new `page-limit` flag has also been added to allow users to specify the number of relationships to ingest per page. This may be useful as a mechanism to throttle the number of relationships read from the datastore.

The new logic detects several scenarios like:
- A backup exists, but no marker exists (meaning it has completed)
- Backup does not exist, but the marker is left behind (it gets truncated)

The marker file will be removed only when the backup completes.

This new logic does not guarantee that relationships will not be duplicated. When terminated gracefully, the system should write the last page received, close, and flush the files. But if the process was abruptly terminated (i.e. SIGKILL) it could lead to relationships being written to the OCF file, but the marker not being updated.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/authzed/zed
 
-go 1.23.8
+go 1.24
 
 toolchain go1.24.1
 
--- a/internal/cmd/backup.go
+++ b/internal/cmd/backup.go
@@ -39,6 +39,16 @@ const (
 	doNotReturnIfExists = false
 )
 
+// cobraRunEFunc is the signature of a cobra.Command.RunE function.
+type cobraRunEFunc = func(cmd *cobra.Command, args []string) (err error)
+
+// withErrorHandling is a wrapper that centralizes error handling, instead of having to scatter it around the command logic.
+func withErrorHandling(f cobraRunEFunc) cobraRunEFunc {
+	return func(cmd *cobra.Command, args []string) (err error) {
+		return addSizeErrInfo(f(cmd, args))
+	}
+}
+
 var (
 	backupCmd = &cobra.Command{
 		Use:   "backup <filename>",
@@ -52,7 +62,7 @@ var (
 		Use:   "create <filename>",
 		Short: "Backup a permission system to a file",
 		Args:  cobra.MaximumNArgs(1),
-		RunE:  backupCreateCmdFunc,
+		RunE:  withErrorHandling(backupCreateCmdFunc),
 	}
 
 	backupRestoreCmd = &cobra.Command{
@@ -314,6 +324,11 @@ func backupCreateCmdFunc(cmd *cobra.Command, args []string) (err error) {
 	}
 
 	defer func(e *error) { *e = errors.Join(*e, encoder.Close()) }(&err)
+
+	if zedToken == nil && cursor == nil {
+		return errors.New("malformed existing backup, consider recreating it")
+	}
+
 	req := &v1.BulkExportRelationshipsRequest{
 		OptionalLimit:  pageLimit,
 		OptionalCursor: cursor,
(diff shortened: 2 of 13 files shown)


internal/cmd/backup.go
```diff
@@ -101,6 +101,8 @@ func registerBackupCmd(rootCmd *cobra.Command) {
 	backupCmd.AddCommand(backupCreateCmd)
 	registerBackupCreateFlags(backupCreateCmd)
 
+	backupCreateCmd.Flags().Uint32("page-limit", 0, "include only schema and relationships with a given prefix")
```
josephschorr, line 109: Fix description

internal/cmd/backup.go
```diff
@@ -337,37 +374,177 @@ func backupCreateCmdFunc(cmd *cobra.Command, args []string) (err error) {
...
 
+// encoderForNewBackup creates a new encoder for a new zed backup file. It returns the ZedToken at which the backup
+// must be taken.
+func encoderForNewBackup(cmd *cobra.Command, c client.Client, backupFile *os.File) (*backupformat.Encoder, *v1.ZedToken, error) {
+	prefixFilter := cobrautil.MustGetString(cmd, "prefix-filter")
+
+	schemaResp, err := c.ReadSchema(cmd.Context(), &v1.ReadSchemaRequest{})
+	if err != nil {
+		return nil, nil, fmt.Errorf("error reading schema: %w", addSizeErrInfo(err))
+	} else if schemaResp.ReadAt == nil {
```
josephschorr, line 427: no need for `} else {` here

internal/cmd/backup.go
```diff
@@ -337,37 +374,177 @@ func backupCreateCmdFunc(cmd *cobra.Command, args []string) (err error) {
...
+
+	return nil
+}
+
+// openProgressFile returns the progress marker file and the stored progress cursor if it exists, or creates
+// a new one if it does not exist. If the backup file exists, but the progress marker does not, it will return an error.
+//
+// The progress marker file keeps track of the last successful cursor received from the server, and is used to resume
+// backups in case of failure.
+func openProgressFile(backupFileName string, backupAlreadyExisted bool) (*os.File, *v1.Cursor, error) {
```
josephschorr, line 481: could we somehow store this in the parquet file itself?

vroldanbet: It's not a parquet file, it's an Avro OCF file, and no: I clarify in the code that OCF are not meant for in-place updates, but for streaming append. I explored it.

internal/cmd/backup.go
vroldanbet: predates the PR, but I'll fix 👍🏻

internal/cmd/backup.go
```diff
@@ -238,71 +255,114 @@ func hasRelPrefix(rel *v1.Relationship, prefix string) bool {
...
-		return fmt.Errorf("unable to initialize client: %w", err)
+		return err
 	}
 
-	ctx := cmd.Context()
-	schemaResp, err := c.ReadSchema(ctx, &v1.ReadSchemaRequest{})
+	defer func(e *error) {
+		*e = errors.Join(*e, backupFile.Sync())
+		*e = errors.Join(*e, backupFile.Close())
+	}(&err)
```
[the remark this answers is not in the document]
vroldanbet: There is no signature in a defer block. The defer block asks you to invoke the function, so it becomes a closure. You can use any shape you want.

alecmerdler (commented):
Not a blocker, but I was able to get an error by running against SpiceDB seeded using the test data:
```
ERR terminated with errors error="error receiving relationships: %!w(<nil>): set flag --max-message-size=9793576 to increase the maximum allowable size
```

vroldanbet:
>Not a blocker, but I was able to get an error by running against SpiceDB seeded using the test data:

@alecmerdler this is a known issue, which can be addressed by setting the `--max-message-size` in case of large schemas, or `--page-limit` (introduced in this PR) to control the size of the pages in a single gRPC message.

I fixed the confusing error message in https://github.com/authzed/zed/pull/492/commits/cea939a4f104ec83c9d8488586e570b4bc696b10, thanks for reporting it!

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/authzed/zed
 
-go 1.23.8
+go 1.24
```
alecmerdler, line 3: Should this be in a separate pull request?

alecmerdler (approved):
Nice fix!

go.mod
vroldanbet: no because I needed it for `t.Context()`

PR #493, opened 2025-04-23
print usage only on parse/validation errors

vroldanbet:
Built on top of https://github.com/authzed/zed/pull/492

The usage was being printed on any error, which was rather unpleasant to look at because it made the output difficult to parse for clues on what happened. The primary goal of this commit is to enhance the user experience.

Cobra supports `SilenceErrors` and `SilenceUsage`, as well as a callback function, `SetFlagErrorFunc`, for handling flag parsing errors and controlling when usage and errors are printed. Unfortunately, only a subset of errors fall under the "flag parsing error" category: validation errors defined via `Command.Args` do not trigger `FlagErrorFunc`, so if `SilenceUsage` is enabled, we would miss the usage for those validation cases 😢

To address this, a special error wrapper is introduced that signals an error is a validation error, allowing the command to manually print the usage when `SilenceUsage` is enabled, which this commit does.

A few tests are added that assert the output. Due to the pervasive use of globals, some refactoring was conducted in `cmd.go` to make it more test-friendly.

This is what the command line output looked like when a normal application error happened:
```sh

```shell
$ zed preview schema compile root.zed
Usage:
  zed preview schema compile <file> [flags]

Examples:

        Write to stdout:
                zed preview schema compile root.zed
        Write to an output file:
                zed preview schema compile --out compiled.zed
        

Flags:
  -h, --help         help for compile
      --out string   output filepath; omitting writes to stdout

Global Flags:
      --certificate-path string     path to certificate authority used to verify secure connections
      --endpoint string             spicedb gRPC API endpoint
      --hostname-override string    override the hostname used in the connection to the endpoint
      --insecure                    connect over a plaintext connection
      --log-format string           format of logs ("auto", "console", "json") (default "auto")
      --log-level string            verbosity of logging ("trace", "debug", "info", "warn", "error") (default "info")
      --max-message-size int        maximum size *in bytes* (defaults to 4_194_304 bytes ~= 4MB) of a gRPC message that can be sent or received by zed
      --no-verify-ca                do not attempt to verify the server's certificate chain and host name
      --permissions-system string   permissions system to query
      --request-id string           optional id to send along with SpiceDB requests for tracing
      --skip-version-check          if true, no version check is performed against the server
      --token string                token used to authenticate to SpiceDB

3:29PM ERR terminated with errors error="failed to read schema file: open root.zed: no such file or directory"
```

This is what it looks like now after the change
```shell

```shell
$ zed preview schema compile root.zed
3:29PM ERR terminated with errors error="failed to read schema file: open root.zed: no such file or directory"
```

### DIFF
--- a/internal/cmd/cmd_test.go
+++ b/internal/cmd/cmd_test.go
@@ -1,0 +1,114 @@
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/jzelinskie/cobrautil/v2/cobrazerolog"
+	"github.com/rs/zerolog"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/require"
+)
+
+// note: these tests mess with global variables, so do not run in parallel with other tests.
+func TestCommandOutput(t *testing.T) {
+	cases := []struct {
+		name                  string
+		flagErrorContains     string
+		expectUsageContains   string
+		expectFlagErrorCalled bool
+		expectStdErrorMsg     string
+		command               []string
+	}{
+		{
+			name:                  "prints usage on invalid command error",
+			command:               []string{"zed", "madeupcommand"},
+			expectFlagErrorCalled: true,
+			flagErrorContains:     "unknown command",
+			expectUsageContains:   "zed [command]",
+		},
+		{
+			name:                  "prints usage on invalid flag error",
+			command:               []string{"zed", "version", "--madeupflag"},
+			expectFlagErrorCalled: true,
+			flagErrorContains:     "unknown flag: --madeupflag",
+			expectUsageContains:   "zed version [flags]",
+		},
+		{
+			name:                  "prints usage on parameter validation error",
+			command:               []string{"zed", "validate"},
+			expectFlagErrorCalled: true,
+			flagErrorContains:     "requires at least 1 arg(s), only received 0",
+			expectUsageContains:   "zed validate <validation_file_or_schema_file> [flags]",
+		},
+		{
+			name:                  "prints correct usage",
+			command:               []string{"zed", "perm", "check"},
+			expectFlagErrorCalled: true,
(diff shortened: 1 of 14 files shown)


internal/cmd/cmd_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/google/uuid"
+	"github.com/jzelinskie/cobrautil/v2/cobrazerolog"
+	"github.com/rs/zerolog"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/require"
+)
+
+// note: these tests mess with global variables, so do not run in parallel with other tests.
+func TestCommandOutput(t *testing.T) {
+	t.Skip()
```
miparnisari, line 17: Is this intentional?

miparnisari (commented):
Could you post examples of how the CLI looks before & after this PR?

internal/cmd/cmd_test.go
vroldanbet: ouch, no

ecordell:
I love the general idea of this, unfortunately I think this makes the UX a little worse in some other cases.


1. It seems to have lost the logger formatting on the error
2. It seems to be losing some context for the specific command usage and not displaying it, it now gives less info than before

Here's a common one I run into with zed (because I can never remember the `check` format - is it one string relation? 5 or 6 fields, one for each tuple component? nope - it's 3 for some reason).

Here's before this PR:
```
$ zed perm check
Usage:
  zed permission check <resource:id> <permission> <subject:id> [flags]

<rest of usage omitted for brevity>

10:52AM ERR terminated with errors error="accepts 3 arg(s), received 0"
```

and after:
```
$ zed perm check
accepts 3 arg(s), received 0
Usage:
  zed [command]

<rest of usage omitted for brevity>

Use "zed [command] --help" for more information about a command.
```

Note that prior to this PR it would give me the info I need - the format to put a check query.

ecordell (approved):
LGTM

PR #494, opened 2025-04-23
add test to validate cmd and fix regression

miparnisari:
Regression introduced in https://github.com/authzed/zed/pull/491


tstirrat15 (approved):
LGTM

PR #495, opened 2025-04-23
fix positioning of error when schema isn't first section

miparnisari:
Before & after

Screenshot 2025-04-23 at 1 04 40 PM


tstirrat15 (approved):
Nice

PR #525, opened 2025-06-26
Update go.mod to retract v1.14.0 

ensonic:
See https://pkg.go.dev/github.com/authzed/zed?tab=versions
+ https://go.dev/blog/go116-module-changes#TOC_5.


tstirrat15 (approved):
Thank you!

PR #530, opened 2025-07-08
validationfile: bubble up validation err with more context

### LINKED ISSUE #529: Make sure that calling `zed validate` on a validation file preserves line number context in the validated schema
When you use `zed validate` with a validation file and with the `validationFile` key, if there's a warning or error in the schema validation, the output doesn't indicate the correct line number. We should fix that.

kartikaysaxena:
fixes #529


kartikaysaxena:
PTAL @tstirrat15

PR #534, opened 2025-08-25
fix(permission): LookupResource should stop the stream according to --page-limit argument

mazdakb:
`zed` calls the LookupResource in a loop, and automatically provides the cursor token. This causes the `--page-limit` to have practically no effect from user's perspective.


tstirrat15 (approved):
Nice. Thank you!

tstirrat15:
Talked to mazdak; sounds like we might go with a slightly different approach.

Issue #552, opened 2025-10-01
Update version command to look at both headers and trailers


tstirrat15:
See authzed/spicedb#2577

Sending the version metadata as a header breaks retry policies (see description in that PR). That means that we moved version metadata from the header to the trailer, which means that the zed logic that looks for the version metadata needs to look in both places to account for different versions of SpiceDB that put it in different places.

We'll need to add the logic and then test against both SpiceDB v1.45.4 (the last version without it) and v1.46.0 (the first version with it).

PR #560, opened 2025-10-04
fix: Update version command to look at both headers and trailers

miparnisari:
Fixes https://github.com/authzed/zed/issues/552


internal/cmd/version.go
```diff
@@ -18,45 +18,59 @@ import (
...
-		fmt.Print(blue("service: "))
-		if len(version) == 1 {
-			console.Println(version[0])
-		} else {
-			console.Println("(unknown)")
+			blue := color.FgLightBlue.Render
+			fmt.Print(blue("service: "))
+			console.Println(serverVersion)
 		}
 	}
```
miparnisari, line 76: I refactored this code because of this warning that my IDE gave me. I think I am preserving behavior.

image

tstirrat15: Why would it always be true? 🤔

tstirrat15: Or is it `hasContext` that's always true by the time you hit this branch?

tstirrat15: nvm I looked at the file without the diff and the logic makes sense to me.

miparnisari: > Why would it always be true? 🤔

`hasContext=true` only in line 31, and by that point, we know that `includeRemoteVersion` is true

PR #568, opened 2025-10-21
Prevent `zed backup create` from swallowing network errors

tstirrat15:
## Description
We got a couple of reports of this. The errors in question were `PermissionDenied` errors, but we were effectively swallowing them because I missed returning the `err` when I refactored this logic in #543.

## Changes
Will annotate

## Testing
Review. See that tests still pass.

### DIFF
--- a/internal/cmd/backup.go
+++ b/internal/cmd/backup.go
@@ -928,5 +928,5 @@ func addPermissionDeniedErrInfo(err error) error {
 	if code != codes.PermissionDenied {
 		return err
 	}
-	return fmt.Errorf("%w: ensure that the token used for this call has all requisite permissions", err)
+	return fmt.Errorf("%w: ensure that the token used for this call has all required permissions", err)
 }


internal/cmd/backup.go
```diff
@@ -398,6 +408,9 @@ func backupCreateCmdFunc(cmd *cobra.Command, args []string) (err error) {
 		}
 		return nil
 	})
+	if err != nil {
+		return err
+	}
```
tstirrat15: This was the substantive change. In #543 I refactored things to use `takeBackup`, which meant that the `return`s that moved into `takeBackup` no longer interrupted execution. I missed returning `err` here because I assumed that it was at the bottom; `ineffassign` didn't yell because of the `defer` blocks that take a pointer to `err`.

internal/cmd/backup.go
```diff
@@ -42,10 +42,20 @@ const (
...
+
 // withErrorHandling is a wrapper that centralizes error handling, instead of having to scatter it around the command logic.
 func withErrorHandling(f cobraRunEFunc) cobraRunEFunc {
 	return func(cmd *cobra.Command, args []string) (err error) {
-		return addSizeErrInfo(f(cmd, args))
+		cmdErr := f(cmd, args)
+		for _, handler := range errorHandlers {
+			cmdErr = handler(cmdErr)
+		}
+		return cmdErr
```
tstirrat15: Changing this to allow us to add other handlers

internal/cmd/backup.go
```diff
@@ -902,3 +915,15 @@ func addSizeErrInfo(err error) error {
...
+func addPermissionDeniedErrInfo(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	code := status.Code(err)
+	if code != codes.PermissionDenied {
+		return err
+	}
+	return fmt.Errorf("%w: ensure that the token used for this call has all requisite permissions")
```
tstirrat15: This should help guide users to setting the appropriate permissions

internal/cmd/backup_test.go
```diff
@@ -281,9 +281,13 @@ func TestBackupCreateCmdFunc(t *testing.T) {
...
 	ctx := t.Context()
 	srv := zedtesting.NewTestServer(ctx, t)
+	runErrs := make(chan error, 1)
 	go func() {
-		require.NoError(t, srv.Run(ctx))
+		err := srv.Run(ctx)
+		runErrs <- err
 	}()
+	err := <-runErrs
+	require.NoError(t, err)
```
tstirrat15: These two refactors are `testifyrequire` lint fixes that I did as drive-bys. We don't have that linter enabled yet. I'd like to do that soon.

PR #574, opened 2025-11-03
build(deps): bump github.com/ccoveille/go-safecast to v2.0.0

ccoVeille:
## Description

I'm the owner of https://github.com/ccoVeille/go-safecast

I noticed your CI was broken because of my recent changes to add deprecation warning to some methods.

- https://github.com/ccoVeille/go-safecast/pull/122

Note: I made some slight changes, as I update the returned value in the v2 of my lib
- https://github.com/ccoVeille/go-safecast/pull/120

So I added check on the error to use zero as a fall back to replicate what would do the v1.

Working on your PR helped me to figure things, such as there is a real need for me to support uintptr

- https://github.com/ccoVeille/go-safecast/issues/17

See https://github.com/authzed/zed/blob/022f31a1892a8612e78ab2a5e902b56a99602539/internal/cmd/schema.go#L226

## Testing

Launched golangci-lint locally.

Please note, I checked that the gosec issue false positive is still there.

## References

Related #573: fix the issue with linting reporting go-safecast deprecation warning of ToInt

- #573 

Related to 
- https://github.com/ccoVeille/go-safecast/issues/126
-  https://github.com/ccoVeille/go-safecast/pull/122
- https://github.com/authzed/spicedb/pull/2685

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/authzed/grpcutil v0.0.0-20240123194739-2ea1e3d2d98b
 	github.com/authzed/spicedb v1.45.4
 	github.com/brianvoe/gofakeit/v6 v6.28.0
-	github.com/ccoveille/go-safecast v1.6.1
+	github.com/ccoveille/go-safecast/v2 v2.0.0
 	github.com/cenkalti/backoff/v4 v4.3.0
 	github.com/charmbracelet/lipgloss v1.1.0
 	github.com/charmbracelet/x/term v0.2.1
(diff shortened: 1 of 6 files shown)


ccoVeille:
recheck

ccoVeille:

recheck

tstirrat15:
Thank you for putting this up! We've already put through a couple of update PRs in our other repos - I like the API changes, and it only broke CI because we're pretty aggressive about deprecation warnings. It would have thrown at some point or another.

And yeah I don't mind having a manual workaround for `uintptr` - the particular functionality is a weird edge case in the golang stdlib and I don't feel like your lib needs to handle it if it'd be too far out of the normal codepath.

tstirrat15 (approved):
Thank you for putting this through!

PR #588, opened 2025-12-01
feat: add --extra-header flag for custom gRPC headers

sputnikraket:
Signed-off-by: Erik Hennings <erik.hennings@freda.com>

## Description
This PR adds support for passing custom headers to gRPC requests via a new `--extra-header` flag.

## Testing

Manual testing

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -21,6 +21,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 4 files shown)


radhus:
> I think this is fine, but what's the use case? I'm wondering if it's something that zed already supports.

We're running SpiceDB behind a proxy that requires authentication (GCP Cloud Run), that can be specified either through the `Authorization` header or another header. So we kind-of need to pass double authentication tokens 😅 

This setup is perhaps not ideal, but it works.

tstirrat15 (commented):
See question; otherwise LGTM

PR #589, opened 2025-12-02
build: more automation on docs workflow

miparnisari:


.github/workflows/docs.yaml
```diff
@@ -37,13 +37,52 @@ jobs:
           ref: "main"
 
       - name: "Sync documentation changes"
+        id: "check-changes"
         run: |
+          set -e # Have script exit in the event of a failed command.
```
squat: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell
This is already the default if the `shell` field is unspecified.
If we want something even more robust, we could set `shell: bash` and get `-o pipefail` for free

PR #597, opened 2025-12-11
feat: add ability to backup systems that don't expose BulkExport API

jzelinskie:
## Description

This PR adds backup support for systems without the bulk export. This is important for allowing folks to perform backups against AuthZed Serverless (as well as old SpiceDB deployments or ones where their API access is restricted).

I went about this change by refactoring a bit of logic in the create backup command and leaning into the Encoder abstraction in the `backupfmt` package. Now the logic for persisting to a file and tracking progress on the CLI are isolated and implemented as their own Encoders.

I didn't try to deduplicate any logic between the bulk export and legacy backup conditional because I expect us to be able to just delete the legacy backup block once AuthZed Serverless is fully sunset.

I did also change how we storing our cursor in the lock file. Previously we held a file descriptor open and kept rewriting it and now I create a new tempfile and atomically replace the lockfile when there's an update; this strategy is more robust to filesystem errors and means we have less bookkeeping.

As a follow-up, I also refactored how we were rewriting legacy prefixes, filtering prefixes, and added support for replacing prefixes by creating a Rewriter interface. This ensures the flags are also the same across commands.

## Testing

I've manually ran this against some permission systems in AuthZed Serverless and written some basic unit tests for the encoder logic that was refactored.

### DIFF
--- a/pkg/backupformat/encoder_test.go
+++ b/pkg/backupformat/encoder_test.go
@@ -416,8 +416,8 @@ func TestNewFileEncoder(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			enc, existed, err := NewFileEncoder(tt.filename)
-			if !existed {
-				enc.WriteSchema(tt.schema, tt.token.Token)
+			if !existed && err == nil {
+				err = enc.WriteSchema(tt.schema, tt.token.Token)
 			}
 			if tt.expectError {
 				require.Error(t, err)
--- a/internal/cmd/backup.go
+++ b/internal/cmd/backup.go
@@ -531,7 +531,7 @@ func backupParseRevisionCmdFunc(cmd *cobra.Command, out io.Writer, args []string
 
 	loadedToken, err := decoder.ZedToken()
 	if loadedToken == nil || err != nil {
-		return fmt.Errorf("failed to parse decoded revision")
+		return errors.New("failed to parse decoded revision")
 	}
 
 	_, err = fmt.Fprintln(out, loadedToken.Token)
--- a/pkg/backupformat/encoder.go
+++ b/pkg/backupformat/encoder.go
@@ -202,7 +202,7 @@ func (fe *OcfFileEncoder) lockFileName() string {
 func (fe *OcfFileEncoder) Cursor() (string, error) {
 	cursorBytes, err := os.ReadFile(fe.lockFileName())
 	if os.IsNotExist(err) {
-		return "", fmt.Errorf("completed backup file already exists")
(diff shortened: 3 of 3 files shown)


internal/cmd/helpers_test.go
```diff
@@ -17,7 +17,7 @@ import (
 func mapRelationshipTuplesToCLIOutput(t *testing.T, input []string) []string {
 	t.Helper()
 
-	return lo.Map[string, string](input, func(item string, _ int) string {
```
tstirrat15: Is there a linter that we could enable for this repo that would catch this?

tstirrat15 (dismissed):
This looks good to me; I'd also like to get maria's eyes on it.

pkg/backupformat/encoder_test.go
```diff
@@ -0,0 +1,555 @@
...
+func TestNewEncoderForExisting(t *testing.T) {
+	var buf bytes.Buffer
+	enc, err := NewEncoderForExisting(&buf)
+
+	require.NoError(t, err)
+	require.NotNil(t, enc)
+	require.NotNil(t, enc.enc)
+}
+
+func TestOcfEncoder_Append(t *testing.T) {
```
miparnisari: can we have similar tests that test `OcfFileEncoder.Append()`?

pkg/backupformat/decoder.go
```diff
@@ -51,34 +51,72 @@ func NewDecoder(r io.Reader) (*Decoder, error) {
...
 
-func (d *Decoder) Next() (*v1.Relationship, error) {
+func (d *OcfDecoder) Schema() (string, error)         { return d.schema, nil }
+func (d *OcfDecoder) ZedToken() (*v1.ZedToken, error) { return d.zedToken, nil }
+func (d *OcfDecoder) Close() error                    { return nil }
+
+func (d *OcfDecoder) Next() (*v1.Relationship, error) {
 	if !d.dec.HasNext() {
-		return nil, nil
+		return nil, io.EOF
```
tstirrat15: Yep, that is weird

pkg/backupformat/redaction.go
```diff
@@ -71,25 +71,33 @@ func (rm RedactionMap) Invert() RedactionMap {
 }
 
 // NewRedactor creates a new redactor that will redact the data as it is written.
-func NewRedactor(dec *Decoder, w io.Writer, opts RedactionOptions) (*Redactor, error) {
+func NewRedactor(dec Decoder, w io.Writer, opts RedactionOptions) (*Redactor, error) {
```
tstirrat15: Why was the pointer change necessary here?

PR #600, opened 2025-12-16
chore: pull in lint rules from spicedb

tstirrat15:
## Description
Makes linting in this repo stricter and brings it in line with what we use on SpiceDB.

This also separates out some of the changes from #597.

## Changes
Will annotate.

## Testing
Review. See that tests pass.

### DIFF
--- a/internal/cmd/schema_test.go
+++ b/internal/cmd/schema_test.go
@@ -128,65 +128,69 @@ caveat test/some_caveat(someCondition int) {
 
 func TestSchemaCompile(t *testing.T) {
 	t.Parallel()
+	require := require.New(t)
 
-	testCases := map[string]struct {
-		files     []string
-		out       string
-		expectErr error
-		expectStr string
-	}{
-		`file_not_found`: {
-			files: []string{
-				filepath.Join("preview-test", "nonexistent.zed"),
-			},
-			expectErr: fs.ErrNotExist,
-		},
-		`happy_path`: {
-			files: []string{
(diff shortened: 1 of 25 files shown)


miparnisari (approved):

internal/cmd/schema_test.go
```diff
@@ -128,65 +128,69 @@ caveat test/some_caveat(someCondition int) {
 
 func TestSchemaCompile(t *testing.T) {
```
tstirrat15: I refactored these to be three separate tests because 1. you can't really use `require.ErrorAs` with a tabular test, at least in its current form (which may change in golang 1.26 but we're not there yet) and 2. one of the error assertions needed to be `ErrorIs`, not `ErrorAs`. There's a bit of duplicated code but I'm fine with that.

internal/cmd/schema_test.go
```diff
@@ -128,65 +128,69 @@ caveat test/some_caveat(someCondition int) {
...
+	mockTermCheckerr := &mockTermChecker{returnVal: false}
+	err := schemaCompileCmdFunc(cmd, files, mockTermCheckerr)
+	require.Error(err)
+	require.ErrorAs(err, &expectedErr)
 }
 
+// TODO: refactor the impl function to provide a pipe or buffer directly and delegate the input selection to
+// another function
+//
+//nolint:tparallel // these tests can't be parallel because they muck around with the definition of os.Stdin.
```
miparnisari: but you kept the `t.parallel` in line 195

Issue #612, opened 2025-12-29
Can't pipe the output of `zed schema compile` to `zed schema write`


julianleonard123:
When using composable schema command to build a schema and write it in a CICD pipeline I can't chain two commands together like this:

```
zed schema compile root.zed | zed schema write
{"level":"error","error":"must provide stdout or output file path","time":"2025-12-29T13:01:06+01:00","message":"terminated with errors"}
1:01PM ERR terminated with errors error="attempted to write empty schema"
```

As a workaround I can redirect the output to a file and run a new command to write the schema, but ideally the default `zed schema compile` which outputs to stdout would be usable via a pipe like shown.

Perhaps if there was a way to specify `--stdout` in the compile command it would work?  Currently it seems zed is expecting a filename with (`--out` flag) or the end of the command.

PR #613, opened 2025-12-29
fix: schema compilation (#612)

miparnisari:
Now the command works correctly in all scenarios:
  1. `zed schema compile file.zed` → writes to stdout
  1. `zed schema compile file.zed 1> out.txt` → writes to redirected stdout
  1. `zed schema compile file.zed --out out.txt` → writes to specified file

Previously, number 2 didn't work.

### DIFF
--- a/internal/cmd/schema.go
+++ b/internal/cmd/schema.go
@@ -83,12 +83,15 @@ func registerAdditionalSchemaCmds(schemaCmd *cobra.Command) *cobra.Command {
 		Example: `
 	Write to stdout:
 		zed preview schema compile root.zed
-	Write to an output file:
+	Write to redirected stdout:
+		zed preview schema compile schema.zed 1> compiled.zed
+	Write to a file:
 		zed preview schema compile root.zed --out compiled.zed
 	`,
 		ValidArgsFunction: commands.FileExtensionCompletions("zed"),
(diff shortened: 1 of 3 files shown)


internal/cmd/schema.go
```diff
@@ -360,22 +363,38 @@ func determinePrefixForSchema(ctx context.Context, specifiedPrefix string, clien
...
-// and produces it as a fully-realized schema
-func schemaCompileCmdFunc(cmd *cobra.Command, args []string, termChecker termChecker) error {
-	stdOutFd, err := safecast.Convert[int](os.Stdout.Fd())
-	if err != nil {
-		return err
-	}
+func schemaCompileOuter(cmd *cobra.Command, args []string) (bool, error) {
 	outputFilepath := cobrautil.MustGetString(cmd, "out")
-	if outputFilepath == "" && !termChecker.IsTerminal(stdOutFd) {
-		return errors.New("must provide stdout or output file path")
```
tstirrat15: If we don't have this do we get weird behavior in environments that don't have stdout available? Or is that captured by the error behavior of writing to `os.Stdout`?

miparnisari: I think it's captured by the call to `fmt.Fprint`.

PR #635, opened 2026-03-02
chore: bring in spicedb and update compilation

tstirrat15:
## Description
This PR is part of unifying the composable schema package into the tooling for a better developer experience. The key changes are:                                  
                                                                                                                                                                    
  1. **Dependency update** — Updates SpiceDB dependency, bringing in a newer version of the schema compilation library.                                  
  2. **Decoder refactor** (internal/decode/decoder.go) — Significant rewrite of the schema decoder, along with updated tests. This simplifies the decoding logic.                                                                                                                                                     
  3. **Validate command simplification** (internal/cmd/validate.go) — Major simplification of the validation logic, cutting roughly half the code. Tests were also reworked                                                                                                                                           
  4. **New tests** — Added 8 new .zed and .yaml test files for edge cases like:                                                                                 
    - Import errors in composable schemas                                                                                                                             
    - Nonexistent imports                                                                                                                                             
    - Nested composable schemas                                                                                                                                       
    - External schemas with escaping/errors  

## Testing
Unit tests


tstirrat15:
Current TODOs and notes:

* This code depends on authzed/spicedb#2965. It should be run locally against that with `go.work`, and that PR should also be tested against `spicedb-vscode`.
- [x] Merge authzed/spicedb#2965 and then `go get` that SHA to make this reference that code.
- [x] Validation tests are failing. My best guess is that one if not all of the failures have to do with the `DirFS()` that's being handed to the DevContext in the validate command - it needs to match the dir of the validation file, not the invocation dir.
- [x] Need to check whether validations against just schemas are pointing their warnings at the correct lines (this should be captured in a test, but should also be validated)
- [x] It needs a test of a schema that imports another schema and there's a validation warning/error in that other schema to see whether it points at the correct file and line within that file
- [x] This should be hand-tested against a few things as well - validation files in the example repo, some composable schema stuff, etc.

tstirrat15 (commented):
See comments

tstirrat15 (commented):
See comments

miparnisari (approved):

PR #672, opened 2026-04-09
chore: get ready to retract versions

miparnisari:
thanks to https://github.com/golang/go/issues/74411#issuecomment-3015012663

i think i now how to mark v1.0.0 as the true latest. We will publish a tag for v1.14.1. This will trigger a release, but that's okay, we'll just delete the release.


tstirrat15 (approved):
LGTM

PR #682, opened 2026-04-25
feat: add lookup resources debugging behavior

tstirrat15:
## Description
Zed counterpart to authzed/spicedb#3070. This does the work of submitting the debug arg and then interpreting the trace held in the error message.

## What the output looks like
```
The following resource/relation pairs were found in a cycle in LookupResources:

	- folder:a#view

To further debug this, issue a check from the resource to itself across the relation.
For example, with the identified pair `resource:foo#view`, you would make the following call:

	zed permission check resource:foo view resource:foo --explain
```

## Changes
Will annotate.

## Testing
Review. `go work` this together with the code in authzed/spicedb#3070 and run the tests. Note that this is blocked on that PR, so tests will fail until that SHA is pulled into this PR.

You can also use the schema and relationships from the tests for a running example - do `go run ./cmd/spicedb/... serve-testing` in one window and then go-run the `lookup-resources` call with `--debug` to see the output.

Alternatively you can `zed import` this:

```yaml

schema: |-
  definition user {}

  definition folder {
    relation viewer: user
    relation parent: folder
    permission view = parent->view + viewer
  }

  definition resource {
    relation folder: folder
    permission view = folder->view
  }
relationships: |-
  folder:a#viewer@user:someuser
  folder:a#parent@folder:b
  folder:b#parent@folder:a
```

And then run the LR for `someuser`


internal/commands/permission.go
```diff
@@ -133,6 +138,7 @@ func RegisterPermissionCmd(rootCmd *cobra.Command) *cobra.Command {
 	lookupResourcesCmd.Flags().Uint32("page-limit", 0, "limit of relations returned per page")
 	lookupResourcesCmd.Flags().String("cursor", "", "resume pagination from a specific cursor token")
 	lookupResourcesCmd.Flags().Bool("show-cursor", true, "display the cursor token after pagination")
+	lookupResourcesCmd.Flags().Bool("debug", false, "send debug header to spicedb. useful for debugging recursion depth errors")
```
tstirrat15, line 141: I'd welcome bikeshedding on this flag. It feels like it's different from `--explain` because it's about getting more context out of an exception rather than showing the path taken on a successful request.

internal/commands/permission.go
```diff
@@ -430,6 +436,11 @@ func lookupResourcesCmdFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	ctx := cmd.Context()
+	if cobrautil.MustGetBool(cmd, "debug") {
+		log.Info().Msg("debugging requested on LookupResources")
+		ctx = requestmeta.AddRequestHeaders(ctx, requestmeta.RequestDebugInformation)
```
tstirrat15: I'd also welcome bikeshedding on this point - I reached for a header by reflex but it may not be how we want to do things.

internal/commands/permission.go
```diff
@@ -503,6 +514,37 @@ func lookupResourcesCmdFunc(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// handleLookupResourcesErr does any additional handling of LR errors.
+// Primarily used to print a traceback when in debug mode.
+func handleLookupResourcesErr(err error) error {
```
tstirrat15, line 520: This feels like it might be worth pushing up into `authzed-go` - unpacking errors always feels like a pain point of gRPC.

tstirrat15 (commented):
See comments

internal/printers/debug.go
```diff
@@ -193,3 +196,16 @@ func isPartOfCycle(checkTrace *v1.CheckDebugTrace, encountered map[string]struct
...
+	output.WriteString("\nThe following resource/relation pairs were found in a cycle in LookupResources:\n\n")
+	for _, member := range debugInfo.CycleMembers {
+		fmt.Fprintf(&output, "\t- %s\n", member)
+	}
+	output.WriteString("\nTo further debug this, issue a check from the resource to itself across the relation.\n")
+	output.WriteString("For example, with the identified pair `resource:foo#view`, you would make the following call:\n")
+	output.WriteString("\n\tzed permission check resource:foo view resource:foo --explain\n\n")
+	output.WriteString("For more information, see the LookupResources section under https://spicedb.dev/d/debug-max-depth\n\n")
+	console.Error(output.String())
+}
```
miparnisari, line 218:
```suggestion
func DisplayLookupResourcesDebugInfo(debugInfo *dispatchv1.LookupDebugInfo) {
	if debugInfo == nil || len(debugInfo.CycleMembers) == 0 {
		return
	}
	parts := strings.Split(debugInfo.CycleMembers[0], "#")
	object := parts[0]
	rel := parts[1]
	var output strings.Builder
	output.WriteString("\nThe following resource/relation pairs were found in a cycle in LookupResources:\n\n")
	for _, member := range debugInfo.CycleMembers {
		fmt.Fprintf(&output, "\t- %s\n", member)
	}
	output.WriteString("\nYou will need to delete a relationship to break the cycle.")
	output.WriteString("\nTo find the pairs involved in the cycle, issue a check from the resource to itself across the relation.\n")
	output.WriteString("For example:\n")
	output.WriteString(fmt.Sprintf("\n\tzed permission check %s %s %s --explain\n\n", object, rel, object))
	output.WriteString("For more information, see the LookupResources section under https://spicedb.dev/d/debug-max-depth\n\n")
	console.Error(output.String())
}
```

miparnisari (dismissed):

miparnisari (approved):

PR #683, opened 2026-04-28
feat: add cursor and limit support to relationship read

kbrwn:
Adds `--cursor` and `--show-cursor` flags to `zed relationship read` so users can resume pagination from a prior cursor token and surface the trailing cursor for follow-up calls.

### DIFF
--- a/internal/commands/relationship.go
+++ b/internal/commands/relationship.go
@@ -107,7 +107,7 @@ func RegisterRelationshipCmd(rootCmd *cobra.Command) *cobra.Command {
 	readCmd.Flags().Bool("json", false, "output as JSON")
 	readCmd.Flags().String("subject-filter", "", "optional subject filter")
 	readCmd.Flags().Uint32("page-limit", 100, "number of relationships queried in each batch when making a no-limit call. used to tune impact on SpiceDB. overridden by --limit when provided")
-	readCmd.Flags().Uint32("limit", 0, "number of relationships returned in a single request. overrides --page-limit when both are provided.")
+	readCmd.Flags().Uint32("limit", 0, "number of relationships returned in total. overrides --page-limit when both are provided.")
 	readCmd.Flags().String("cursor", "", "resume pagination from a specific cursor token")
 	readCmd.Flags().Bool("show-cursor", false, "display the cursor token after pagination")
 	registerConsistencyFlags(readCmd.Flags())
--- a/internal/commands/relationship.go
+++ b/internal/commands/relationship.go
@@ -107,6 +107,8 @@ func RegisterRelationshipCmd(rootCmd *cobra.Command) *cobra.Command {
 	readCmd.Flags().Bool("json", false, "output as JSON")
 	readCmd.Flags().String("subject-filter", "", "optional subject filter")
 	readCmd.Flags().Uint32("page-limit", 100, "limit of relations returned per page")
+	readCmd.Flags().String("cursor", "", "resume pagination from a specific cursor token")
+	readCmd.Flags().Bool("show-cursor", false, "display the cursor token after pagination")
 	registerConsistencyFlags(readCmd.Flags())
 
 	relationshipCmd.AddCommand(bulkDeleteCmd)
@@ -278,7 +280,10 @@ func readRelationships(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	lastCursor := request.OptionalCursor
+	var lastCursor *v1.Cursor
+	if cursorStr := cobrautil.MustGetString(cmd, "cursor"); cursorStr != "" {
+		lastCursor = &v1.Cursor{Token: cursorStr}
+	}
 	for {
 		request.OptionalCursor = lastCursor
 		var cursorToken string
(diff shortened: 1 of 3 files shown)


internal/commands/relationship.go
```diff
@@ -259,6 +261,10 @@ func buildRelationshipsFilter(cmd *cobra.Command, args []string) (*v1.Relationsh
 }
 
 func readRelationships(cmd *cobra.Command, args []string) error {
+	return readRelationshipsImpl(cmd, args, printRelationship)
+}
+
+func readRelationshipsImpl(cmd *cobra.Command, args []string, responseHandler func(cmd *cobra.Command, responses []*v1.ReadRelationshipsResponse) error) error {
```
tstirrat15, line 268: Splitting off the impl so that we can provide a callback in tests and not have to mock thigns

internal/commands/relationship.go
```diff
@@ -278,7 +284,10 @@ func readRelationships(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	lastCursor := request.OptionalCursor
+	var lastCursor *v1.Cursor
+	if cursorStr := cobrautil.MustGetString(cmd, "cursor"); cursorStr != "" {
+		lastCursor = &v1.Cursor{Token: cursorStr}
+	}
```
tstirrat15, line 297: This is the input of the page

internal/commands/relationship.go
```diff
@@ -306,38 +316,47 @@ func readRelationships(cmd *cobra.Command, args []string) error {
...
+			responses = append(responses, msg)
 			lastCursor = msg.AfterResultCursor
 			relCount++
-			if err := printRelationship(cmd, msg); err != nil {
-				return err
-			}
+		}
+
+		if err := responseHandler(cmd, responses); err != nil {
+			return err
```
tstirrat15, line 332: Call handler with pages, not with individual requests; this lets us test page behavior without meaningfully changing the handling logic.

internal/commands/relationship.go
```diff
@@ -306,38 +316,47 @@ func readRelationships(cmd *cobra.Command, args []string) error {
...
 }
 
-func printRelationship(cmd *cobra.Command, msg *v1.ReadRelationshipsResponse) error {
-	if cobrautil.MustGetBool(cmd, "json") {
-		prettyProto, err := PrettyProto(msg)
-		if err != nil {
-			return err
-		}
+func printRelationship(cmd *cobra.Command, responses []*v1.ReadRelationshipsResponse) error {
+	for _, response := range responses {
```
tstirrat15, line 354: This change is just making it iterate over the page.

internal/commands/relationship_test.go
```diff
@@ -647,8 +649,7 @@ func (m *mockClient) WriteRelationships(_ context.Context, in *v1.WriteRelations
 }
 
 func TestBulkDeleteForcing(t *testing.T) {
-	ctx, cancel := context.WithCancel(t.Context())
-	defer cancel()
+	ctx := t.Context()
```
tstirrat15, line 652: drive-by

internal/commands/relationship_test.go
```diff
@@ -820,3 +819,152 @@ func assertRelationshipCount(ctx context.Context, t *testing.T, c client.Client,
...
+// makeTestCallback is a convenience function that returns a callback and a pointer to
+// the slice it modifies.
+func makeTestCallback() (func(_ *cobra.Command, responses []*v1.ReadRelationshipsResponse) error, *[][]*v1.ReadRelationshipsResponse) {
+	var pages [][]*v1.ReadRelationshipsResponse
+	callback := func(_ *cobra.Command, responses []*v1.ReadRelationshipsResponse) error {
+		pages = append(pages, responses)
+		return nil
+	}
+	return callback, &pages
+}
```
tstirrat15, line 848: This is both a little awkward and maybe a little cute. The big thing is that it needs to return a pointer to the slice rather than the slice itself in order for the callback to correctly modify it, but then you have to deref it where it's used.

It's not used in *that* many places so I can inline if desired.

tstirrat15 (commented):
See comments

internal/commands/relationship.go
```diff
@@ -106,7 +106,10 @@ func RegisterRelationshipCmd(rootCmd *cobra.Command) *cobra.Command {
 	relationshipCmd.AddCommand(readCmd)
 	readCmd.Flags().Bool("json", false, "output as JSON")
 	readCmd.Flags().String("subject-filter", "", "optional subject filter")
-	readCmd.Flags().Uint32("page-limit", 100, "limit of relations returned per page")
+	readCmd.Flags().Uint32("page-limit", 100, "number of relationships queried in each batch when making a no-limit call. used to tune impact on SpiceDB. overridden by --limit when provided")
+	readCmd.Flags().Uint32("limit", 0, "number of relationships returned in a single request. overrides --page-limit when both are provided.")
```
tstirrat15: Here's where we add the concrete `limit`

internal/commands/relationship.go
```diff
@@ -271,14 +278,23 @@ func readRelationships(cmd *cobra.Command, args []string) error {
...
 	request := &v1.ReadRelationshipsRequest{RelationshipFilter: filter}
 
-	limit := cobrautil.MustGetUint32(cmd, "page-limit")
+	pageSize := cobrautil.MustGetUint32(cmd, "page-limit")
+	limit := cobrautil.MustGetUint32(cmd, "limit")
+	limitSet := limit > 0
+	// If --limit is not set, use --page-size as the per-request limit
+	if !limitSet {
+		limit = pageSize
+	}
```
tstirrat15, line 287: This is where we decide whether we have a limit and what it should be.

internal/commands/relationship.go
```diff
@@ -306,38 +323,49 @@ func readRelationships(cmd *cobra.Command, args []string) error {
...
 
-		if relCount < limit || limit == 0 {
-			return nil
+		if err := responseHandler(cmd, responses); err != nil {
+			return err
+		}
+
+		// If we set an explicit limit or if we find a page smaller than the page limit,
+		// we stop.
+		if limitSet || relCount < pageSize {
```
tstirrat15: This logic changed - we stop iterating if we've explicitly set a limit, or else if the relCount is smaller than the limit.

internal/commands/relationship.go
miparnisari:
```suggestion
	readCmd.Flags().Uint32("limit", 0, "number of relationships returned in total. overrides --page-limit when both are provided.")
```

miparnisari (dismissed):

alecmerdler (dismissed):

alecmerdler (approved):

PR #688, opened 2026-05-04
fix: write default backup to cwd and print saved path

### LINKED ISSUE #679: `zed import` is defaulting to storing the backup in the folder where the `zed` binary lives
```
[12:54:12] ~/Desktop/zed-demo (master) $ zed backup create
12:54PM WRN not calling a released version of SpiceDB version=a81358557206-dirty
⠋ processing backup (71018/-, 751325 relationship/s) [0s]

[12:54:19] ~/Desktop/zed-demo (master) $ find /opt/homebrew /usr/local "$HOME" -name "*.zedbackup" 2>/dev/null
/opt/homebrew/bin/local.zedbackup
```

1. `zed backup create` should be printing out where the filename got stored (the full path). Especially if I didn't specify one 😄 
2. we should not be using the location of the binary but the current working directory

ivanauth:
## Summary

Closes #679.

- `zed backup create` (with no filename arg) now writes the `.zedbackup` to the current working directory instead of the directory of the `zed` binary.
- Both `zed backup create` and `zed backup redact` now print `Backup written to <abs-path>` on stderr after a successful close, so users can see where the file ended up.
- Drive-by: `backupRedactCmdFunc` switched to a named return `(err error)` so the existing `defer CloseAndJoin(&err, ...)` calls actually propagate close errors to the caller (they were previously mutating a shadowed local).

### DIFF
--- a/internal/cmd/backup.go
+++ b/internal/cmd/backup.go
@@ -242,6 +242,17 @@ func backupCreateCmdFunc(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
+	// Defer the success message so it runs only after the encoder has been
+	// closed without error. Defers run LIFO, so this fires after the encoder
+	// close below has had a chance to mutate err. The completed flag prevents
+	// printing during panic unwinding.
+	completed := false
+	defer func() {
+		if completed && err == nil {
+			printBackupFileLocation(backupFileName)
+		}
+	}()
+
 	spiceClient, err := client.NewClient(cmd)
 	if err != nil {
 		return fmt.Errorf("unable to initialize client: %w", err)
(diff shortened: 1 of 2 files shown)


miparnisari (approved):
Thank you! Tested locally and works great.

internal/cmd/backup.go
```diff
@@ -262,7 +273,24 @@ func backupCreateCmdFunc(cmd *cobra.Command, args []string) (err error) {
...
+		return err
+	}
+	completed = true
+	return nil
+}
+
+// printBackupFileLocation prints the absolute path of the backup file to stderr,
+// unless the destination is stdout (i.e. "-").
+func printBackupFileLocation(filename string) {
+	if filename == "-" {
```
miparnisari, line 286: TIL. Could you add tests for this? It was introduced in https://github.com/authzed/zed/pull/285/files and I think at some point we broke it, because I get

```go
$ ./zed backup create - --log-level trace
12:13PM DBG configured logging async=false format=auto log_level=trace provider=zerolog
12:13PM TRC context={"APIToken":"super-secret-key","CACert":null,"Endpoint":"localhost:50051","HostnameOverride":"","Insecure":true,"Name":"local","NoVerifyCA":false} context-override-via-cli=false
12:13PM WRN not calling a released version of SpiceDB version=97811cf1455d-dirty
12:13PM TRC parsed trailers trailers={"io.spicedb.debug.version":["97811cf1455d-dirty"],"io.spicedb.respmeta.cachedoperationscount":["0"],"io.spicedb.respmeta.dispatchedoperationscount":["1"],"x-request-id":["7ac1f5c6-92b2-42b2-926f-ae5192c9dd38"]}
12:13PM DBG extracted response dispatch metadata cached=0 dispatch=1
// ommitted for brevity

412:13PM ERR terminated with errors error="failed to store cursor in lockfile: cannot create temp file: open /dev/stdout.lock343325193: operation not permitted\nsync /dev/stdout: inappropriate ioctl for device"
```

PR #691, opened 2026-05-04
fix: support backup create when streaming to stdout

### LINKED ISSUE #284: zed backup to stdout or existing file

ivanauth:
## Summary

`zed backup create -` (write backup to stdout) currently fails with:

```
failed to store cursor in lockfile: cannot create temp file: open /dev/stdout.lock343325193: operation not permitted
sync /dev/stdout: inappropriate ioctl for device
```

`OcfFileEncoder` was attempting to write a cursor lockfile derived from `fe.file.Name()` (resolving to `/dev/stdout.lock`) and calling `file.Sync()` / `file.Close()` on `os.Stdout` during shutdown.

This was reported in the review of #688: https://github.com/authzed/zed/pull/688#pullrequestreview-4222773540.

Closes #284.

### DIFF
--- a/docs/zed.md
+++ b/docs/zed.md
@@ -84,7 +84,7 @@ zed backup <filename> [flags]
 
 ### Children commands
 
-- [zed backup create](#reference-zed-backup-create)	 - Backup a permission system to a file
+- [zed backup create](#reference-zed-backup-create)	 - Back up a permission system to a file, or to stdout by passing "-"
 - [zed backup parse-relationships](#reference-zed-backup-parse-relationships)	 - Extract the relationships from a backup file
 - [zed backup parse-revision](#reference-zed-backup-parse-revision)	 - Extract the revision from a backup file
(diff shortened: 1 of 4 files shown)


miparnisari (commented):
Do you mind updating the help text of `zed backup create`? Currently it says

```
$ ./zed backup create --help
Backup a permission system to a file

Usage:
  zed backup create <filename> [flags]
```

let's make it "Backup a permission system to a file or to stdout (denoted by a dash)"

and `zed backup create <filename-or-dash>`

ivanauth:
Updated in ecc773a.

miparnisari (approved):

PR #699, opened 2026-06-11
fix(cli): support '-' for stdin in backup restore

### LINKED ISSUE #698: zed backup restore fails when using '-' for stdin
### Description

Currently, `zed backup restore` fails when attempting to restore from stdin using the standard `-` argument. The command terminates with the following error:
`unable to stat restore file: stat -: no such file or directory`

While piping data without arguments works when stdin is not a terminal, explicitly passing `-` is a standard Unix convention that currently fails because `openRestoreFile` attempts to call `os.Stat()` on the literal string `"-"`.

This should match the existing behavior of `zed backup create -` (implemented in #688), which handles the dash shorthand correctly.

### Steps to Reproduce
```bash
cat backup | zed backup restore -

Juqsi:
## Description

`zed backup restore -` currently fails with `unable to stat restore file: stat -: no such file or directory` because the code attempts to call `os.Stat()` on the literal `"-"` string.

This PR adds standard Unix `-` shorthand support for stdin, aligning the behavior with `zed backup create -` (implemented in #688). `openRestoreFile` now treats `filename == "-"` identically to `filename == ""`, returning `os.Stdin` directly.

## Testing

* Added table-driven unit test `TestOpenRestoreFile` in `backup_test.go` covering `""`, `-`, valid file paths, and invalid file errors
* Manually verified via: `cat backup.zed | zed backup restore -`

## References

* Closes #698
* Relates to #688

### DIFF
--- a/internal/cmd/backup.go
+++ b/internal/cmd/backup.go
@@ -474,7 +474,7 @@ func isCanceled(err error) bool {
 }
 
 func openRestoreFile(filename string) (*os.File, int64, error) {
-	if filename == "" {
+	if filename == "" || filename == "-" {
 		log.Trace().Str("filename", "(stdin)").Send()
 		return os.Stdin, -1, nil
 	}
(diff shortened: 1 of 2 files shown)


Juqsi:

Juqsi:
recheck

PR #722, opened 2026-09-04
chore: improve argument handling and error messages

tstirrat15:
Written with LLM assistance.

## Description
I ran into this when I fatfingered a `zed permission lookup-resources` today. The error looks like this:

```
❯ zed permission lookup-resources thumper/user:someuser view thumper/resource
5:56PM ERR terminated with errors error="the length of the unpacked is not equal to the provided input"
```

Which is cryptic and doesn't indicate where the issue is - it sounds like it could as easily be a gRPC internal error as a malformed input error.

This PR improves the error output so that it points at the specific arg and provides a template.

## Changes
* Add more parsing helpers and improve existing parsing helpers
* Use those parsing helpers
* Tests for parsing helpers
* Test of overall command

## Testing
Review. See that tests pass. do `go run ./cmd/... permission lookup-resources thumper/user:someuser view thumper/resource` and see that you get more useful output.


internal/commands/permission.go
```diff
@@ -546,15 +548,17 @@ func handleLookupResourcesErr(err error) error {
...
 		return err
 	}
 
 	permission := args[1]
 
-	subjectType, subjectRelation := ParseType(args[2])
+	subjectType, subjectRelation, err := ParseType(args[2])
+	if err != nil {
+		return err
+	}
```
miparnisari, line 561: can you add a test covering LS?

miparnisari (dismissed):

miparnisari (approved):

